PR #3, opened 2018-10-22
Parameterized SMTP details and admin password

gs11:
Fixes https://github.com/gravitee-io/issues/issues/1600

### DIFF
--- a/gravitee/values.yaml
+++ b/gravitee/values.yaml
@@ -25,6 +25,20 @@ oauth:
 inMemoryAuth:
   enabled: true
 
+# Default password "admin", use bcrypt to generate a new one
+adminPasswordBcrypt: $2a$10$Ihk05VSds5rUSgMdsMVi9OKMIx2yUvMz7y9VP3rJmQeizZLrhLMyq
+
+smtp:
+  enabled: true
+  host: smtp.example.com
+  port: 25
+  from: info@example.com
+  username: info@example.com
+  password: example.com
+  properties:
+    auth: true
+    starttlsEnable: false
+
 mongo:
   rs: rs
   rsEnabled: true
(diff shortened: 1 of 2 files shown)


gravitee/values.yaml
```diff
@@ -25,6 +25,19 @@ oauth:
 inMemoryAuth:
   enabled: true
 
+# Default password "admin", use bcrypt to generate a new one
+adminPasswordBcrypt: aaa
```
brasseld: I think it would be useful to keep the default value from `gravitee.yml` here.

brasseld:
```suggestion
adminPasswordBcrypt: $2a$10$Ihk05VSds5rUSgMdsMVi9OKMIx2yUvMz7y9VP3rJmQeizZLrhLMyq
```

gravitee/values.yaml
```diff
@@ -25,6 +25,19 @@ oauth:
...
+
+smtp:
+  host: smtp.example.com
```
brasseld: Why not also providing the item for `enabled` property ?

brasseld:
```suggestion
  enabled: true
  host: smtp.example.com
```

gravitee/values.yaml
gs11: Sorry, leftover from testing. Fixed.

gravitee/values.yaml
gs11: Fixed

gs11:
All suggested changes implemented, thanks for reviewing!

brasseld:
Please squash your commits into a single one.
Once done, I will merge the PR.

gs11:
Done @brasseld

brasseld:
Thanks a lot for your contribution!

PR #8, opened 2019-03-20
Gateway debug logging flag

patst:
I just had the need to enable debug logging for the gateway to solve a problem.

Unfortunately this was not possible with the chart but it is possible for the api deployment.

I added the necessary code to enable the debug logs for the gateway with the `gateway.debugEnabled`  flag.

### DIFF
--- a/gravitee/README.md
+++ b/gravitee/README.md
@@ -121,7 +121,7 @@ To configure common functionalities such as:
 | Parameter             | Description       | Default   |
 |-----------------------|-------------------|-----------|
 | `api.name`            | API service name  | `api`     |
-| `api.debugEnabled`    | Whether to enable API debug or not  | `false`     |
+| `api.logging.debug`    | Whether to enable API debug logging or not  | `false`     |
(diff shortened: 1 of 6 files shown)


brasseld:
Hi @patst 

May I ask you to rename the property to something like `gateway.logging.debug` instead of `gateway.debugEnabled` just to be clear with the fact that we are enabling debug mode for logs, and not the gateway itself.

And we surely have to apply the same thing for the management-api. Perhaps you can handle it also ?

patst:
Hey @brasseld 

thanks for the reply. Sounds like a good idea. I can change it for the management-api too.
What do you think about introducing something like `gateway.logging.level` and use the value in the logback.xml to have the flexibility to switch between different log levels?

patst:
> Hi @patst
> 
> May I ask you to rename the property to something like `gateway.logging.debug` instead of `gateway.debugEnabled` just to be clear with the fact that we are enabling debug mode for logs, and not the gateway itself.
> 
> And we surely have to apply the same thing for the management-api. Perhaps you can handle it also ?

I changed the parameter for the gateway and management api to `logging.debug` and adjusted the readme as well.

brasseld:
Thanks a lot @patst!

PR #9, opened 2019-03-20
Use variable for gateway ingress path

patst:
We want to use a different host for the gateway ingress together with the path `/` instead of the default `/gateway`.

I added a new property (`gateway.ingress.path`) which defaults to the current default. (`/gateway`)


brasseld:
In the same way as the other PR, may I ask you to apply the same for both the management-api and ui please ?

patst:
Sure, I made the changes for the management API as well and added the new properties `gateway.ingress.path` and `api.ingress.path` in the helm charts readme.

brasseld:
Thanks a lot @patst!

PR #10, opened 2019-03-20
Fix broken ingress path variable

patst:
With https://github.com/gravitee-io/gravitee-kubernetes/pull/9 the helm chart got messed up, because inside the `range $host` part we cannot reference the variable `ingress.path` directly.

@brasseld  sorry for the messup.
I fixed it and got it deployed this way.


brasseld:
Oh yes, perfect!

PR #18, opened 2019-06-17
Node selector fix (it was not working because it was not at the good place)

bloudman:
The node selector was in "containers" it should be in "spec"


brasseld:
Please squash the 3 commits into a single one.

Thanks,

bloudman:
I squashed everything

brasseld:
Thanks a lot!

PR #19, opened 2019-08-02
Allow to define extra in memory users from helm chart values

weseven:
This small PR allows to define additional in memory users from the helm chart values.


brasseld:
Nice!

Thanks a lot

PR #20, opened 2019-08-29
add support for extra volume mounts in ui container

chicco785:
this pr allows to mount extra volumes to ui containers
this is particularly useful for example to load additional themes and locales


chicco785:
@brasseld i noticed that chart version is kind of equal to app version, usually chart versioning in independent.

PR #25, opened 2019-10-10
components optional, k8s 1.16 apis, websocket, sharding tags

previ:
- All components are optionals, so it's possible to use the same chart to deploy the management api and ui separate from the gateway, as well as multiple shards of the gateway in the same cluster.
- websocket support configurable via values parameters
- sharding tags configurable via values parameters


gravitee/README.md
```diff
@@ -174,6 +174,10 @@ See [Elasticsearch](https://github.com/helm/charts/tree/master/stable/elasticsea
...
+| `gateway.sharding_tags`                        | Sharding tags (comma separated list)                                                                                                                                  | ``                                                                                                                                                                                                                          |
+| `gateway.websocket`                            | Whether websocket protocol is enabled or not                                                                                                                          | `false`                                                                                                                                                                                                                     |
```
chicco785: duplicated?

gravitee/README.md
```diff
@@ -174,6 +174,10 @@ See [Elasticsearch](https://github.com/helm/charts/tree/master/stable/elasticsea
...
+| `gateway.sharding_tags`                        | Sharding tags (comma separated list)                                                                                                                                  | ``                                                                                                                                                                                                                          |
```
chicco785: duplicated?

gravitee/values.yaml
```diff
@@ -209,14 +211,18 @@ api:
...
 
 gateway:
+  enabled: true
   type: Deployment
   name: gateway
   logging:
     debug: false
   replicaCount: 2
+  sharding_tags: default
+  tenant:
```
chicco785: no value? no error?

brasseld: I would advise to remove these to lines because if the gateway is started on the default sharing tags, you have to register this one as part of the management api. Without this, you will not be able to deploy apis to gateway

chicco785:
it would be good if you squash the commits when all changes are done, so there is a single commit added ;)

previ:
> it would be good if you squash the commits when all changes are done, so there is a single commit added ;)

Done.

previ:
Hi, is there something to fix/update?

chicco785:
what about @brasseld comment?

previ:
Ah sorry, I didn't see it. I just removed the 'default' value for sharding, which should be equivalent to removing the line.
I am going to comment both lines to avoid any confusion, as suggested.

chicco785:
@brasseld is it ok?

brasseld:
Yes, I was looking for the latest changes, but I did not get any notification for them...

PR #26, opened 2019-10-28
Split repository to manage products from Gravitee.io

brasseld:
Helm chart has been splitted into apim and am to manage products from Gravitee.io


am/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+name: am
+version: 1.0.0
```
chicco785: i would start from a <1.0.0 and move to 1.0.0 when it is tested enough

apim/README.md
```diff
@@ -25,7 +25,7 @@ $ helm package .
 To install the chart using the chart archive, run:
 
 ```
-$ helm install gravitee-1.23.2.tgz
+$ helm install apim-1.27.1.tgz
```
chicco785: did you think about auto generating them and hosting them on github?

chicco785: you can find some example here: https://github.com/orchestracities/charts

brasseld: > did you think about auto generating them and hosting them on github?

Yes, I'm currently working on this but I have some ssl certificates issues with github pages when adding the helm repo:

```shell
helm repo add graviteeio https://helm.gravitee.io
Error: Looks like "https://helm.gravitee.io" is not a valid chart repository or cannot be reached: Get https://helm.gravitee.io/index.yaml: x509: certificate is valid for www.github.com, *.github.io, *.githubusercontent.com, *.github.com, github.com, github.io, githubusercontent.com, not helm.gravitee.io
```

brasseld: I think it works well for you because you're still using a github domain

> helm repo index . --url https://orchestracities.github.io/charts/

am/Chart.yaml
brasseld: Fixed

apim/README.md
chicco785: right, not sure you can do that if not with a redirect . not sure this can help: https://github.blog/2018-05-01-github-pages-custom-domains-https/

brasseld: I already had a look with no luck...

brasseld: Don't know what was the issue with DNS records but seems to be ok now (maybe due to dns propagation?)

PR #28, opened 2019-10-28
changed hard coded service type to variable

dmytrokosiachenko:
Changing service type in values.yml have no result on helm deployment, so this should fix it.


brasseld:
Good catch, thx !

PR #30, opened 2019-11-04
chore(#2835): change deprecated k8s APIs to newer ones

l-lin:
BREAKING CHANGE: use kubernetes v1.16+

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,15 +7,12 @@ The repository includes the following charts:
 - Gravitee.io Access Management (in development)
 - Gravitee.io Alert Engine 
 
-**Note:** These charts are only compatible with Kubernetes < 1.16. For latest versions of k8s, please check that issue: https://github.com/gravitee-io/issues/issues/2835 and the associated Pull Request: https://github.com/gravitee-io/helm-charts/pull/30
-
-
 ## Charts
 
 Please look in the chart directories for the documentation for each chart. These helm charts are designed to be a lightweight way to configure our official docker images
 
 | Chart                                      | Docker documentation                                                            |
 | ------------------------------------------ | ------------------------------------------------------------------------------- |
-| [API Management](./apim/README.md)         | https://docs.gravitee.io/apim/1.x/apim_installguide_kubernetes.html                      |
+| [API Management](./apim/README.md)         | https://docs.gravitee.io/apim/1.x/apim_installguide_kubernetes.html             |
 | [Access Management](./am/README.md)        | https://docs.gravitee.io/am/2.x/am_installguide_docker.html                     |
 | [Alert Engine](./ae/README.md)             | https://docs.gravitee.io/ae/installguide_docker.html                            |
--- a/am/templates/_helpers.tpl
+++ b/am/templates/_helpers.tpl
@@ -53,3 +53,26 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- printf "%s-%s-%s" .Release.Name $name .Values.ui.name | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Return the apiVersion of deployment.
+*/}}
+{{- define "deployment.apiVersion" -}}
+{{- if .Capabilities.APIVersions.Has "apps/v1" -}}
+    {{- print "apps/v1" -}}
+{{- else -}}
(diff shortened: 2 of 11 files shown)


chicco785:
i am not sure how to handle such breaking changes. @brasseld ?

brasseld:
Not so easy to handle, I should also check that most of the major cloud providers are compliant with k8s 1.16+ (GCP: https://cloud.google.com/kubernetes-engine/docs/release-notes#version_updates)

The easiest way would be two manage two branches but I'm not sure about maintainability...

chicco785:
i would then avoid to introduce the change now, and given that 1.16 still handles it, keep it as it is for now

l-lin:
I added templates in the `_helpers.tpl` to set the `apiVersion` according to the Kubernetes apiVersion. So it should not break for old version of Kubernetes.

I used the same technique as [Vault helm chart to support both apiVersion](https://github.com/helm/charts/blob/master/incubator/vault/templates/_helpers.tpl#L45-L65).

brasseld:
Hi @l-lin 

Good catch ! Will have a try asap !

Jean-Baptiste-Lasselle:
hi Gravitee comunity, I want to reference here a case that you might find of interest : 
* I use AWS in a current project 
* I have of course, picked up @l-lin issue, mentioned in the helm-charts `README.md`
* I described [here](https://github.com/gravitee-io/issues/issues/2835#issuecomment-620060810) What happened to me in details
* And I still will confirm, but I am almost sure by now, that AWS releases its own distribution of Kubernetes, and not caring about the why, for example, a `1.14/1.15` Kubernetes in AWS, matches a "real" Kubernetes version above `v1.16`. I know that for sure, because I have the following message in my AWS logs : 
```bash 
    warning: extensions/v1beta1/Ingress is deprecated by networking/v1beta1/Ingress and not supported by Kubernetes v1.20+ clusters.
 ```
* And I have : 
```bash 
# -
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:58:59Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15+", GitVersion:"v1.15.11-eks-af3caf", GitCommit:"af3caf6136cd355f467083651cc1010a499f59b1", GitTreeState:"clean", BuildDate:"2020-03-27T21:51:36Z", GoVersion:"go1.12.17", Compiler:"gc", Platform:"linux/amd64"}
```
* I think there will be numerous future Gravitee users who will use AWS EKS in the future
* And AWS Kubernetes, both versions `1.14` and `1.15` are above `1.16` Kubernetes : which means that in some time soon, those teams won't be able to deploy Gaavitee to production on AWS, deoesn't it ?


And finally, I added those perspective here, for I believe this thread is where work happens, while the issue is a management reference.

Hope this will be valuable, Gravitee rocks ! :)

Jean-Baptiste-Lasselle:
Oh! And also thank u @l-lin for giving the pattern which very interestingly well applied to Vault too !

chopmann:
--> .Values.ui.extraVolumes missing in values.yaml

brasseld:
Hey guys,

Sorry for being late here. I didn't have time to check this PR.
Looks @Jean-Baptiste-Lasselle did the job before me :)

Is this validated only for EKS? Someone checks with GCP / AKS ?

Jean-Baptiste-Lasselle:
> Is this validated only for EKS? Someone checks with GCP / AKS ?

Oh hi @brasseld I am focused on work tasks, and did not check social github . So far : 
* No, I haven't finished my work on my side, so that I cannot give all proof about what I stated regarding `AWS EKS`
* I promise I will give feed back on whether, and how it affects ability to work with GRavitee in Kubernetes Stack, using AWS EKS as Cloud Provider
* Pretty sure I will be able to confirm the reality of the  `EKS` Distribution of Kubernetes

arealmaas:
What is the current status here? We are going to deploy an AKS cluster with kubernetes 1.17 as it becomes GA from 1st July. Would prefer using the official gravitee helm chart instead of basing it on this PR :)

brasseld:
Hello there,

I'm closing the PR as latest charts are now compatible with 1.16.

PR #31, opened 2019-11-08
added support for aws alb ingress controller 

dmytrokosiachenko:
Added support for aws-alb-ingress-controller.
If this changes are too ingress-controller specific, please make suggestion how can support for this ingress-controller be implemented.


chicco785:
i would agree that changes are too ingress specific, e.g. why can't you make configurable the path in the values and load it from there?

dmytrokosiachenko:
Because ui path is not configurable over values.yaml, api and gateway path are used in ui-configmap.yaml so I can't just use /* and for  ssl termination new rule in all three ingresses  in specific order is needed.

PR #32, opened 2019-11-14
Add aws alb ingress controller support

dmytrokosiachenko:
Second try  :)
These changes are needed to provide support for aws-alb-ingress controller.
I can also provide documentation how values.yaml should looks like so that aws alb ingress controller could detect ingresses and how to turn on ssl termination.


apim/templates/ui-configmap.yaml
```diff
@@ -15,15 +15,15 @@ data:
   gzip-types: "*"
   constants.json: |
     {
-      "baseURL": "https://{{ index .Values.api.ingress.hosts 0 }}{{ .Values.api.ingress.path }}/",
+      "baseURL": "https://{{ index .Values.api.ingress.hosts 0 }}/management/",
```
brasseld: Why did you change this part?

apim/templates/ui-configmap.yaml
```diff
@@ -15,15 +15,15 @@ data:
...
       "management": {
         "title": "{{ .Values.ui.managementTitle }}"
       },
       "documentation": {
         "url": "{{ .Values.ui.documentationLink }}"
       },
       "portal": {
-        "entrypoint": "https://{{ index .Values.gateway.ingress.hosts 0 }}{{ .Values.gateway.ingress.path }}",
+        "entrypoint": "https://{{ index .Values.gateway.ingress.hosts 0 }}/gateway",
```
brasseld: Same question here...

apim/templates/ui-deployment.yaml
```diff
@@ -40,7 +40,7 @@ spec:
           imagePullPolicy: {{ .Values.ui.image.pullPolicy }}
           env:
             - name: MGMT_API_URL
-              value: "https://{{index .Values.api.ingress.hosts 0 }}{{ .Values.api.ingress.path }}/"
+              value: "https://{{index .Values.api.ingress.hosts 0 }}/management/"
```
brasseld: Same question here...

apim/templates/ui-configmap.yaml
dmytrokosiachenko: thank you for quick response. 
Wenn all pathes for ingress should be mapped from aws alb ingress controller, there must stay wildcard like  /*  , simple slash is not enough for this ingress controller. It is well known behaviour  see hier -> https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/667
If i change api.ingress.path to /* than baseURL looks like https://api.example.com/*/ , what is not valid path.
Same for other changes.

brasseld: In that case, why not providing two distinct properties:
- gateway.ingress.path (for conf)
- gateway.ingress.exposed-path (for ingress)

WDYT?

@chicco785 any idea?

dmytrokosiachenko: i also thought about this possibility, it can be a little confusing because we don't use than path anywhere in ingress, only in configmaps. In default case ( no aws-alb-ingress-controller) these variables are identical.
But it is definitely better than hard-coded pathes.

Another possibility is to filter out wildcard from path in configmaps with replace.

brasseld: Having a new boolean property `wildcard` under `ingress`, at the same level as `path` would not be sufficient?

If `wildcard` is `true` then, set `/*` for all ingresses path, and use only `path` for configmap.

WDYT?

dmytrokosiachenko: sounds good, but can possibly lead to failures because developer doesn't expect that all endpoints are mapped from ingress controller and that regardless of the fact, that developer configured his path in ingress.apth. What do you think about this filter:
https://github.com/dmytrokosiachenko/gravitee-helm/pull/1/files

dmytrokosiachenko: @brasseld what do you think, please respond

brasseld:
Hey @dmytrokosiachenko 

Do you think this issue still make sense after the latest changes to the chart for APIM?

brasseld:
Closing the PR for inactivity.

PR #36, opened 2019-12-30
Improve APIM configurability

Alexandre-Carbenay:
Add configuration entrypoints for:
- init containers and extra volumes
- pods postStart and preStop scripts
- mongoDB servers & ssl
- Management API logging
- Management API HTTP client
- various Management API & UI keys


brasseld:
Can you squash the two commits into a single one?

Alexandre-Carbenay:
Done

apim/templates/api-configmap.yaml
```diff
@@ -42,8 +88,9 @@ data:
           password: {{ .Values.es.security.password }}
         {{- end }}
         index: {{ .Values.es.index }}
+        cluster: {{ .Values.es.cluster }}
```
brasseld: AFAIK, cluster is useless

PR #37, opened 2020-01-03
Ensure Management API and Gateway pods restart after configuration change

Alexandre-Carbenay:
Based on https://github.com/technosophos/k8s-helm/blob/master/docs/charts_tips_and_tricks.md#automatically-roll-deployments-when-configmaps-or-secrets-change


brasseld:
LGTM

Just one note (in case of): the `checksum/config` name is immaterial and can be subject to change.

PR #40, opened 2020-01-10
Enable SSL activation between ingress and pods

Alexandre-Carbenay:


apim/values.yaml
```diff
@@ -257,6 +257,16 @@ api:
     pullPolicy: Always
   # pullSecrets:
   #   - name: gravitee_secrets
+  ssl:
+    enabled: false
+#    keystore:
+#      path: ${gravitee.home}/security/keystore.jks
+#      password: secret
+    clientAuth:
```
brasseld: I would prefer to keep the same structure we can have in `gravitee.yml` https://github.com/gravitee-io/gravitee-gateway/blob/1.30.x/gravitee-gateway-standalone/gravitee-gateway-standalone-distribution/src/main/resources/config/gravitee.yml#L27

Or maybe you have a good reason to change it?

apim/values.yaml
```diff
@@ -343,6 +353,16 @@ gateway:
   logging:
     debug: false
   reloadOnConfigChange: true
+  ssl:
+    enabled: false
+    #    keystore:
+    #      path: ${gravitee.home}/security/keystore.jks
+    #      password: secret
+    clientAuth:
```
brasseld: Same comment as the previous one.

PR #42, opened 2020-01-13
Improve configurability

mickaelalibert:
## What type of PR is this?

It adds needed configurability for our use case that could be needed by other
This only concerns APIM chart

This PR is a proposal for the usability of the chart, including some new parameters for configurability:
 - Moves chart templates to subfolders, allowing us to apply files via kubectl more precisely in a pipeline 
 - Exposes the [technical api](https://docs.gravitee.io/apim_installguide_management_api_technical_api.html) of the management app for monitoring purposes
 - Updates ingress annotations ( sometimes nginx, sometimes not), now everything is configured for nginx by default, and the patch for nginx [rewrite-target change](https://kubernetes.github.io/ingress-nginx/examples/rewrite/#rewrite-target) in version 0.22.0
 - Updates Readme to document the new configurations values
 - Add some warning about using ingress/app-root with the same host  

None

## Special notes for your reviewer:
First of all, thank you for the time spent creating this project, which was really usefull for our company use case @ RCA. We are currently considering using gravitee, and this is why this PR exists. 
Do not hesitate to get in touch with me if you need any information, or if I can do anything regarding this PR.


brasseld:
Hey @mickaelalibert 

Thanks for this amazing PR.

I'll have a look to it and try it on our platforms and go back to you asap.

Regards,

apim/templates/ui/ui-configmap.yaml
```diff
@@ -29,6 +29,9 @@ data:
       {{- if or .Values.gateway.enabled .Values.ui.title .Values.ui.portal }},
       "portal": {
         {{- if .Values.gateway.enabled }}
+        {{- if .Values.gateway.entrypoint }}
```
brasseld: I'm not sure why we would need this kind of change.

Maybe you can explain it by providing an example and, better, provide some documentation about this property and its usage ?

apim/values.yaml
```diff
@@ -363,19 +380,24 @@ gateway:
     maxReplicas: 6
     targetAverageUtilization: 50
     targetMemoryAverageUtilization: 80
+  # This is usefull in case of dynamic rewriting
+  entrypoint: /gateway/
```
brasseld: The same here.

Can you explain the purpose of this property ?
Do not see the purpose in the case of dyn rewriting, I'm probably missing something there...

apim/values.yaml
```diff
@@ -363,19 +380,24 @@ gateway:
...
       nginx.ingress.kubernetes.io/ssl-redirect: "false"
       nginx.ingress.kubernetes.io/enable-rewrite-log: "true"
-      kubernetes.io/app-root: /gateway
-      kubernetes.io/rewrite-target: /gateway
-      # ingress.kubernetes.io/configuration-snippet: "etag on;\nproxy_pass_header ETag;\nproxy_set_header if-match \"\";\n"
+      # App-root can be used for convenience purpose if
+      # ingress hosts are only used for the gateway on the path '/'
+      # (Value should be similar to ingress.path)
+      # nginx.ingress.kubernetes.io/app-root: /gateway/
+      nginx.ingress.kubernetes.io/rewrite-target: /$1
```
brasseld: So, as far as I understand, it means that when creating API and defining their context-path we don't have to put the `/gateway` anymore, am I right?

And, in that case, it means that the one who was using previous version of the chart will have to update all his API's context-path to make it works.

Please confirm.

apim/templates/ui/ui-configmap.yaml
mickaelalibert: This is due to the way `nginx.ingress.kubernetes.io/rewrite-target` now works (since v0.22.0). 
Before 0.22, this: `nginx.ingress.kubernetes.io/rewrite-target: "/"` means that `GET ingress/toto` forwards to `mysvc/toto`
From 0.22, this:  `nginx.ingress.kubernetes.io/rewrite-target: "/"` means that `GET ingress/toto` forwards to `mysvc/`

This is due to the warning explicited [here](https://kubernetes.github.io/ingress-nginx/examples/rewrite/#rewrite-target) ` In Version 0.22.0 and beyond, any substrings within the request URI that need to be passed to the rewritten path must explicitly be defined in a capture group.`

If your ingress controller is updated, using a matching group in ingress path (as for example `(.*)`) will break the entrypoint value, meaning that any link from the ui to the gateway for `mygreatapi/v1` will look like `https://myapi.ingress.com/gateway/(.*)mygreatapi/v1`

Using `ingress.path` might even be counter indicative for this purpose

apim/values.yaml
mickaelalibert: I could maybe use the description I put in the README `Defines the gateway entrypoint for the portal (In some case, this value is not needed because 'gateway.ingress.path' does not contain matching group and is usable as is for the gateway entrypoint)`
I'll work in the next few days on explaining this more clearly in the README, with some use case if needed

apim/values.yaml
mickaelalibert: You are right, and you are also right this will break things  up. This slipped through my review, as this is something we needed in our use case, but the goal of this PR was to keep changes to a minimum for current users.

mickaelalibert:
@reviewers:
I disabled `nginx.ingress.kubernetes.io/app-root:` for every ingress, as the 'default' values where a little bit counter-intuitive regarding ingress, and the fact that when using app-root, any call to `/` will 302 to the app-root of the first ingress alphabetically when nginx-controller parses configuration. So multiple app-root should not be use with the same host, as this would result in random redirections. 

Regarding my PR as a whole, I don't really know the existing users and their usage, but I think it will be a little easier to understand what is trying to be done regarding redirections and exposure with a dedicated documentation. I could work on that inside this PR if you agreed on the need.

brasseld:
Hi @mickaelalibert 

> Regarding my PR as a whole, I don't really know the existing users and their usage, but I think it will be a little easier to understand what is trying to be done regarding redirections and exposure with a dedicated documentation. I could work on that inside this PR if you agreed on the need.

I agree on the need so it would be perfect if you can update documentation.

Also, do you think it's possible to provide distinct PRs for :
- Moving to subfolders
- Technical API configuration and exposure
- nginx ingress upgrade

Thanks a lot.

brasseld:
Hi @mickaelalibert 

Can you confirm that this PR can be close ?

mickaelalibert:
yes, sorry

PR #43, opened 2020-01-19
:truck: Moving each component to a specific subfolder

mickaelalibert:
## What type of PR is this?

A structural change to the Helm template for the api manager

This PR is a proposal for the usability of the chart:
 - Moves chart templates to subfolders, allowing us to apply files via kubectl more precisely in a pipeline for example 

None


brasseld:
LGTM

Many thanks for the PR.

PR #44, opened 2020-01-19
:sparkles: Expose technical api

mickaelalibert:
## What type of PR is this?
Improvement in configurability

This PR allows to expose the [technical api](https://docs.gravitee.io/apim_installguide_management_api_technical_api.html) of the management app for monitoring purposes

None


apim/templates/api-configmap.yaml
```diff
@@ -48,14 +48,15 @@ data:
     services:
       core:
         http:
-          enabled: true
-          port: 18083
-          host: localhost
+          enabled: {{ .Values.api.http.services.core.http.enabled }}
```
brasseld: You may probably have to manage the default value to `true` for those who are not using the default `values.yml` and providing their own.

brasseld: (I mean, for backward compatibility)

mickaelalibert: I removed the possibility to disabled the technical api, as this is not really needed, and testing a value to be nil/false/true in helm isn't an easy task (cf multiple issues as this one https://github.com/helm/helm/issues/3308)

brasseld:
LGTM.

Please squash your 4 commits into a single one.

PR #47, opened 2020-01-28
Configure emails for in-memory users

Alexandre-Carbenay:


apim/templates/api/api-configmap.yaml
```diff
@@ -114,11 +114,13 @@ data:
           # - none : passwords are not hashed/encrypted
           # default value is bcrypt
           password-encoding-algo: bcrypt
+          allow-email-in-search-results: {{ .Values.inMemoryAuth.allowEmailInSearchResults }}
           users:
             - user:
               username: admin
               password: {{ .Values.adminPasswordBcrypt }}
               roles: MANAGEMENT:ADMIN, PORTAL:ADMIN
+              email: {{ .Values.adminEmail }}
```
brasseld: Do you think that it makes sense here to set the email if `adminEmail` is empty?

Alexandre-Carbenay: No, but this is the default value and adding a condition to just ignore it will reduce readability and have the same effect, so I preferred adding it whatever the value. Would you prefer me to add the condition?

brasseld: Nope, it was just to be sure

Alexandre-Carbenay: So do you need further changes to approve this PR?

apim/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: apim
-version: 1.29.1
+version: 1.30.0
```
brasseld: Can we consider a 1.29.2 version instead please?

The rest of the PR is ok for me.

Alexandre-Carbenay: Changed

brasseld:
👍

PR #49, opened 2020-01-30
Fix typos in docs. for dependency parameters

nafarlee:
The parameters that correspond to enabling the Elasticsearch and Mongo DB dependencies are `elasticsearch.enabled` and `mongodb-replicaset.enabled` respectively, at least according to the `requirements.yaml`. This is not what is outlined in the documentation.


brasseld:
Nice shot !

Thanks

PR #53, opened 2020-02-18
Added the localhost parameters to SMTP configuration

oyviaase:
Our email set-up requires that one of the arguments it receives from the client is it's fully-qualified domain name. These changes provides a method to specify a domain-name that is provided as an argument to the mail server.


brasseld:
Thanks for your contribution!

PR #56, opened 2020-03-11
[FIX] apim: Use correct service values for mongo and elasticsearch

gustavovalverde:
When installing using helm, mongo and elasticsearch services uses the helm name in the README, which is: `graviteeio-apim`


brasseld:
Most of the people are using external Mongo / ES, so that's probably the reason why we missed this...

Thanks a lot!

PR #57, opened 2020-03-11
[FIX] apim: Add nginx to configuration-snippet on ingress annotations

gustavovalverde:
The right annotation is `nginx.ingress.kubernetes.io/configuration-snippet`


brasseld:
Thanks for the contrib @gustavovalverde

PR #58, opened 2020-03-18
Update SSL configuration for gateway and Management API to be compliant with 1.30 version

Alexandre-Carbenay:


Alexandre-Carbenay:
Hello David. Will you have some time soon to validate this PR? Thanks

PR #62, opened 2020-07-06
Add support for openshift routes to the helm chart this Fixes gravitee-io/issues#4062

oyviaase:
Suggestion on how to add support for openshift routes to the helm chart. It's not completely done the labels needs to be fixed as well as some other small touch ups that have chagned since the version of the helm chart this is based on.  Please let me know if there is anything else you would like us to change as well.

### DIFF
--- a/apim/1.x/values.yaml
+++ b/apim/1.x/values.yaml
@@ -8,6 +8,9 @@ inMemoryAuth:
   enabled: true
   allowEmailInSearchResults: false
 
+openshiftRoutes:
+  enabled: false
+
 # Default password "admin", use bcrypt ($2a$ version) to generate a new one
 adminPasswordBcrypt: $2a$10$Ihk05VSds5rUSgMdsMVi9OKMIx2yUvMz7y9VP3rJmQeizZLrhLMyq
 adminEmail:
--- a/apim/1.x/README.md
+++ b/apim/1.x/README.md
@@ -64,11 +64,12 @@ To configure common features such as:
 - logs database (see [elastichsearch](https://github.com/helm/charts/tree/master/stable/elasticsearch)
   chart)
 
-| Parameter              | Description        | Default |
-| ---------------------- | ------------------ | ------- |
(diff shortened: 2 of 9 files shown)


apim/1.x/values.yaml
```diff
@@ -360,6 +360,9 @@ api:
   ingress:
     enabled: true
     path: /management
+    # Set enabled to true to use openshift-routes instead of kubernetes-ingress
+    route:
+      enabled: false
```
brasseld, line 365: You have defined a route flag per component (gateway, api, ui, ...).

I'm wondering if it would not be easier to manage an "openshift mode" with a single / global flag property ?

Also, to make things easier to understand, I would rename `route` for `openshift` to remove ambiguity.

WDYT?

oyviaase: A single "openshift mode" global flag is in my opinion a bad idea, as people are using openshift differently some are using the routes openshift provides while other uses the stock ingress method provided by kubernetes. 

I'd rather have a couple of global openshift flags one for using openshift-routes instead of kubernetes-ingress and perhaps one for security contexts which should probably be added as another PR

oyviaase:
@brasseld can you review the changes?

oyviaase:
Is there any progress on this @brasseld? I've implemented a master switch to enable/disable the use of openshift-routes instead on the traditional kubernetes alternative

oyviaase:
@brasseld can you review the changes? I've tried to fix the merge conflict, but it didn't work. Do you know if I merging your changes from upstream into my repo would solve the merge conflict?

brasseld:
Hi, you have to:
* Merge your commits into a single one (to keep a correct history)
* Merge latests changes from master

oyviaase:
@brasseld I was unable to merge everything down into one commit in this PR so I've opened up a new one where I've done that here: https://github.com/gravitee-io/helm-charts/pull/71 

So I'm just closing this PR

PR #63, opened 2020-07-21
k8s: Ensure running runAsNonRoot for all Gravitee pods

brasseld:
Closes gravitee-io/issues#4078

### DIFF
--- a/apim/3.x/Chart.yaml
+++ b/apim/3.x/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: apim3
-version: 3.0.3
-appVersion: 3.0.4
+version: 3.0.4
+appVersion: 3.1.1
 description: Official Gravitee.io Helm chart for API Management 3.x
 home: https://gravitee.io
 sources:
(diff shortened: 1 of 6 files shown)


MarkDeckert:
I think this will do it.

Just curious about changing user from 1001 to 101.  1001 does work with the previous image for 3.0.4 as long as its defined in the yaml, while 101 does not. There will be less chance of breaking change if user is set to 1001 perhaps.  Otherwise i'm not super opinionated on the choice.  

Thanks for making the change so quickly.  I would have offered a PR myself except that I'm less familiar with helm charts than I should be.

brasseld:
Sorry for being late on this.

I choose 101 instead of 101 just because it's the one used by the nginx-alpine image: 
https://github.com/nginxinc/docker-nginx/blob/master/stable/alpine/Dockerfile#L11

For 3.0.4 you would have to use the previous version of the helm chart, and then move to `3.0.4` version of the helm chart for g.io > 3.1.x, so I would say, at least, no major breaking changes here.

WDYT?

MarkDeckert:
@brasseld This makes sense! I like that decision as it increases simplicity.

PR #64, opened 2020-08-05
Support technical api healthcheck

chicco785:
we experienced that when connection to mongo due to overload get lost, configuration may get lost. it seems that the tcp healtcheck would detect the service as healthy, despite it is not.

hopefully supporting the technical api health check may mitigate the issue


chicco785:
health check on tech api is enabled if `.Values.gateway.http.services.core.service.enabled` and `.Values.api.http.services.core.service.enabled` host need to be set to 0.0.0.0 to be reached by health check of k8s

brasseld:
Hi @chicco785 

I'm having a look to your PR. Do you think you can resolve the conflicts ?

chicco785:
@brasseld conflicts resolved

apim/1.x/templates/gateway/gateway-deployment.yaml
```diff
@@ -81,6 +81,26 @@ spec:
                 command: {{ .Values.gateway.lifecycle.preStop }}
             {{- end }}
           {{- end }}
+          {{- if .Values.gateway.http.services.core.service.enabled }}
+          livenessProbe:
```
brasseld, line 85: I'm a bit concerned by this `livenessProbe`.
The API Gateway is / should be resilient against the MongoDB defects and still able to handle incoming HTTP requests.
By defining this livenessProbe, we are assuming that the API Gateway is not able to perform anymore when MongoDB is down.
Is it the behavior you're looking to see here ?

brasseld (changes requested):

apim/1.x/templates/gateway/gateway-deployment.yaml
chicco785: @brasseld unfortunately this is the behaviour we experience, on a test cluster we have during load test experienced that if timeouts with mongodb occurs a few times (not necessarily all together), gateway loses configuration and does not recovers it. this results that apis are not reachable. still for the tcp health check all is good. only solution is to restart the affected gateway manually. so we are looking for this restart to be made automatically if needed.

brasseld: > on a test cluster we have during load test experienced that if timeouts with mongodb occurs a few times (not necessarily all together), gateway loses configuration and does not recovers it.

I'm a bit surprised by that one. Once loaded from Mongo to the Gateway, this later do not really need Mongo anymore. That's the reason why I don't understand you said that the gateway is affected.

The only thing, according to me, would be that the health endpoint for technical API is provided a failing probe result which makes the gateway unreachable from the ingress and does not accept new request anymore.

chicco785: then i am not sure it is mongodb the problem, what isn't the technical api health checking that everything is ok?

`GET /_node/health`

> Get the health status of the component. If the return status is 200, then everything’s ok, if 500, there’s at least one probe in error. This endpoint could be used by your loadbalancer to know if a component instance could be out of the pool.

`HTTP/1.1 200 OK
Content-Type: application/json

{
    "management-api": {
        "healthy": true
    },
    "management-repository": {
        "healthy": true
    },
    "repository-analytics": {
        "healthy": true
    }
}`

brasseld: Yes, for me that is the problem.

If the health endpoint returns a 500 because of a probe failed, the ingress controller will remove the pod to accept requests.

But the gateway is still up and can continue to handle requests, even if MongoDB is down.

chicco785: well, as said, maybe then the problem is not mongo, but still the issue occurs, and this health check detects the problem, and restarts the pod and everything gets back to normal

chicco785: @brasseld ?

brasseld: Hi @chicco785 

I tried to reproduce your problem by overloading the gateway.
Even if it ends with MongoDB timeout, API are still accessible through the gateway.
Can you tell me with which version of Gravitee.io you are experiencing this issue?

chicco785: problem showed up with 1.23.1, updates to up 1.30 didn't seem to change the issue, we had to change mongo configuration to a better resource reservation

PR #70, opened 2020-10-21
Enable or disable elasticsearch from values

ydbondt:


apim/3.x/templates/api/api-configmap.yaml
```diff
@@ -116,6 +116,7 @@ data:
     analytics:
       type: elasticsearch
       elasticsearch:
+        enabled: {{ .Values.es.enabled }}
```
ydbondt, line 119: Apparently this can't be disabled in api config ?

brasseld: Hi @ydbondt 

For now, analytics can't be disable for mgmt-api.
You don't want analytics ?

brasseld (changes requested):
Looking for feedback

apim/3.x/templates/api/api-configmap.yaml
ydbondt: It is not that we don't want analytics, but when we do a proof of concept we don't want to be forced to install a rather big component like elasticsearch.

arne-heil:
Please implement this feature. We could really use a minimal footprint with at least Elasticsearch as an optional choice. Thanks in advance :smile:

brasseld:
> Please implement this feature. We could really use a minimal footprint with at least Elasticsearch as an optional choice. Thanks in advance 😄

Hi @arne-heil 

Currently, Elasticsearch is an optional dependency for the API Gateway (you could rely on other reporters like TCP, file, ...).
But it's still not at the Management API level.

Why disabling ES would need a minimal footprint (for the Management API) ?

arne-heil:
> > Please implement this feature. We could really use a minimal footprint with at least Elasticsearch as an optional choice. Thanks in advance smile
> 
> Hi @arne-heil
> 
> Currently, Elasticsearch is an optional dependency for the API Gateway (you could rely on other reporters like TCP, file, ...).
> But it's still not at the Management API level.
> 
> Why disabling ES would need a minimal footprint (for the Management API) ?

We would like to enable and disable ES when needed. Otherwise we have to maintain a ES instance only for Gravitee usage and/or disable the Management API after setting up and deploying our API.

brasseld:
> We would like to enable and disable ES when needed. Otherwise we have to maintain a ES instance only for Gravitee usage and/or disable the Management API after setting up and deploying our API.

Meaning that you're happy not having any kind of analytics from Gravitee, right?

arne-heil:
> > We would like to enable and disable ES when needed. Otherwise we have to maintain a ES instance only for Gravitee usage and/or disable the Management API after setting up and deploying our API.
> 
> Meaning that you're happy not having any kind of analytics from Gravitee, right?

In general, yes. We don't have any complains about Gravitee in combination with our API  and analytics would be a nice optional feature for us.

ydbondt:
If there is any intention to apply this, you can reopen the pull request.

PR #71, opened 2020-11-05
Resubmition of: Add support for openshift routes to the helm chart this Fixes gravitee-io/issues#4062

oyviaase:
Resubmitting from a new branch as I was unable to merge everything into one commit in the master branch

### DIFF
--- a/apim/1.x/README.md
+++ b/apim/1.x/README.md
@@ -64,11 +64,12 @@ To configure common features such as:
 - logs database (see [elastichsearch](https://github.com/helm/charts/tree/master/stable/elasticsearch)
   chart)
 
-| Parameter              | Description        | Default |
-| ---------------------- | ------------------ | ------- |
-| `chaos.enabled`        | Enable Chaos test  | false   |
-| `inMemoryAuth.enabled` | Enable oauth login | true    |
-| `ldap.enabled`         | Enable LDAP login  | false   |
+| Parameter                 | Description        | Default |
+| ------------------------- | ------------------ | ------- |
+| `chaos.enabled`           | Enable Chaos test  | false   |
+| `inMemoryAuth.enabled`    | Enable oauth login | true    |
+| `ldap.enabled`            | Enable LDAP login  | false   |
+| `openshiftRoutes.enabled` | Use openshift-route instead of ingress | false |
 
 ### Mongo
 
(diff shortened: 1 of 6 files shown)


apim/1.x/README.md
```diff
@@ -64,11 +64,12 @@ To configure common features such as:
...
-| ---------------------- | ------------------ | ------- |
-| `chaos.enabled`        | Enable Chaos test  | false   |
-| `inMemoryAuth.enabled` | Enable oauth login | true    |
-| `ldap.enabled`         | Enable LDAP login  | false   |
+| Parameter                 | Description        | Default |
+| ------------------------- | ------------------ | ------- |
+| `chaos.enabled`           | Enable Chaos test  | false   |
+| `inMemoryAuth.enabled`    | Enable oauth login | true    |
+| `ldap.enabled`            | Enable LDAP login  | false   |
+| `openshiftRoutes.enabled` | Use openshift-route instead of ingress | false |
```
brasseld, line 72: Hi @oyviaase 

Could it be `openshift.enabled` instead ?
I'm assuming maybe more changes in the future because of an "openshift mode" so I think it could be better to not manage openshift routes specifically.

WDYT ?

brasseld (changes requested):

apim/1.x/README.md
oyviaase: Hi @brasseld 

I'd rather still have a seperate option to enable or disable openshift routes, as there might exists users running gravitee on openshift that's using ingress instead off routes. 

So what about having `openshit.enabled` to controll the use of a future "openshift mode" and then under that key add `openshift.route.enabled` to controll weather or not to use openshift routes?

WDYT?

brasseld: Thanks for your feedback.
Unfortunately, I don't know so much about OpenShift. Does it mean that users who are deploying in OpenShift can rely on classical ingresses instead of routes ?

oyviaase: Yes they can, OpenShift afik fully support classical ingress. Users of openshift might still want to use routes instead as they find benefits with that solution over ingress. 

For a quick overview of routes VS ingress scroll down to the table near the end of this article https://www.openshift.com/blog/kubernetes-ingress-vs-openshift-route

oyviaase:
This was for apim1

PR #72, opened 2020-11-23
Allow the configuration of an OIDC-Provider via Helm-chart

Andy2003:
resolves https://github.com/gravitee-io/issues/issues/4645

### DIFF
--- a/apim/3.x/templates/api/api-configmap.yaml
+++ b/apim/3.x/templates/api/api-configmap.yaml
@@ -141,6 +141,9 @@ data:
         index: {{ .Values.es.index }}
     security:
       providers:
+        {{- if and .Values.security.providers }}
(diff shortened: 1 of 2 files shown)


brasseld:
Hi @Andy2003 

Thanks for the PR. 

I'm just wondering how can we manage array of oidc here.

Any thoughts ?

Andy2003:
we could move the whole yaml out into the values, so we just insert the yaml as is. But this will remove the ability to simple configure the chart via parameters

Andy2003:
@brasseld Any opinion on my [commet](https://github.com/gravitee-io/helm-charts/pull/72#issuecomment-732254201)

brasseld:
Sorry for being late on that one. I do agree with you.

I'll check the PR asap.

Andy2003:
@brasseld I have added a 2nd configuration option, with which you can also configure multiple providers

PR #74, opened 2020-12-08
feat: add custom management console URL handling

jbouchery:
resolves #4739

### DIFF
--- a/apim/3.x/templates/api/api-configmap.yaml
+++ b/apim/3.x/templates/api/api-configmap.yaml
@@ -34,6 +34,9 @@ data:
     {{- end }}
 
     management:
+      {{- if .Values.ui.ingress.enabled }}
+      url: https://{{ index .Values.ui.ingress.hosts 0 }}{{ regexFind "/[a-zA-Z]+" .Values.ui.ingress.path }}/
+      {{- end }}
       type: {{ .Values.management.type | default "mongodb" }}
       {{- if or (eq .Values.management.type "mongodb") (kindIs "invalid" .Values.management.type) }}
       mongodb:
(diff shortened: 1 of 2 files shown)


apim/3.x/templates/api/api-configmap.yaml
```diff
@@ -34,6 +34,9 @@ data:
     {{- end }}
 
     management:
+      {{- if .Values.ui.ingress.enabled }}
+      url: https://{{ index .Values.ui.ingress.hosts 0 }}{{ regexFind "/[a-zA-Z]+" .Values.ui.ingress.path }}/
```
brasseld, line 38: what is the purpose of the regexFind here ?

jbouchery: By default, the path is /console(/|$)(.*) and this regexFind keep the /console part (Keep the "/" and n letters)

Maybe i must also add numbers.

brasseld: Can't we simply keep `.Values.ui.ingress.path` only without the trailing slash ?

jbouchery: Yes but actually the default ingress rule doesn't to work without the trailing slash, even if i use this configuration : https://kubernetes.github.io/ingress-nginx/examples/rewrite/#rewrite-target

I have to type `https://my.awesome.url/console/` to work, without the trailing slash every call go to  `https://my.awesome.url/` (so JS and CSS calls don't go to my console pod)

brasseld: Yes, I still don't understand this, and had no time to waste on it.

PR #76, opened 2020-12-18
add condition for replica setting

Durburz:
Helm3 replaces the replica count set by the HPA repeatedly, resulting in a up- and down-scaling loop.

Because of that it is necessary to update the helm charts.

### DIFF
--- a/ae/Chart.yaml
+++ b/ae/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: ae
-version: 1.1.12
+version: 1.1.13
 appVersion: 1.2.18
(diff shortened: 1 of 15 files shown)


brasseld:
Thanks for the PR! 
👍

modularTaco:
Hey @brasseld ,

I can see that this PR increased the version from `1.29.13` to `1.29.14`. But the latest available version of the apim1 chart in the Gravitee Helm repository (https://helm.gravitee.io/index.yaml) is still `1.29.13` and is missing the fixes from this PR.

Is there any reason why this has been merged, but not published to the Helm repository?

PR #79, opened 2021-01-19
[#4922] Fix mongo env var overloading using helm

Samaruq:
This commit changes the mongo section of the gateway configmap template.
The section is updated to use the same format as the api configmap
template. This allows for mongo parameter overloading using env vars.


Samaruq:
https://github.com/gravitee-io/issues/issues/4922

brasseld:
Thanks a lot @Samaruq 👍

PR #81, opened 2021-01-22
Fix requirements - Update Helm Chart Repository

seluard:
Since November 13, 2020, the new location for the stable repository is https://charts.helm.sh/stable and the new location for the incubator repository is https://charts.helm.sh/incubator.

It was change on apim, but not in am and designer.

J.L.


brasseld:
Thanks a lot @seluard

PR #82, opened 2021-01-27
Mtls socket nginx annotation on ingress

Jean-Baptiste-Lasselle:
see https://github.com/gravitee-io/gravitee-cockpit/issues/142 : 

I think what here makes sense, is that we add this annotation directly into the source code of the Helm Chart : that's about MLTS and cockpit connecting with APIM and AM, even in the source code of Cockpit, this is not an option, this is a core feature of Cockpit. @brasseld agreee ?


brasseld:
👍

PR #85, opened 2021-02-02
Update ui-deployment.yaml

pciatos:
Removed ending '-', this one was leading the following error when ui.runAsNonRoot == true:
Error: YAML parse error on am/templates/ui/ui-deployment.yaml: error converting YAML to JSON: yaml: line 35: mapping values are not allowed in this context
helm.go:75: [debug] error converting YAML to JSON: yaml: line 35: mapping values are not allowed in this context
YAML parse error on am/templates/ui/ui-deployment.yaml


brasseld:
👍

PR #87, opened 2021-02-05
[#2820] add mongo authSource field to configmaps when not using Mongo URI

Samaruq:
`.Values.mongo.auth.source` is only used if mongo uri is used. This PR allows to set the `authSource` in the resulting `gravitee.yml` with this value if mongo uri is not used.

Related to gravitee-io/issues#2820.

### DIFF
--- a/apim/3.x/templates/api/api-configmap.yaml
+++ b/apim/3.x/templates/api/api-configmap.yaml
@@ -51,6 +51,7 @@ data:
         {{- if (eq .Values.mongo.auth.enabled true) }}
         username: {{ .Values.mongo.auth.username }}
         password: {{ .Values.mongo.auth.password }}
+        authSource: {{ .Values.mongo.auth.source | default "gravitee" }}
         {{- end }}
         {{- else }}
         uri: mongodb://{{- if (eq .Values.mongo.auth.enabled true) }}{{ .Values.mongo.auth.username }}:{{ .Values.mongo.auth.password }}@{{- end }}{{ .Values.mongo.dbhost }}:{{ .Values.mongo.dbport }}/{{ .Values.mongo.dbname }}?{{- if .Values.mongo.rsEnabled }}&replicaSet={{ .Values.mongo.rs }}{{- end }}{{- if (eq .Values.mongo.auth.enabled true) }}&authSource={{ .Values.mongo.auth.source }}{{- end }}{{- if .Values.mongo.connectTimeoutMS }}&connectTimeoutMS={{ .Values.mongo.connectTimeoutMS }}{{- end }}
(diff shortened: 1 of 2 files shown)


apim/3.x/templates/api/api-configmap.yaml
```diff
@@ -51,6 +51,7 @@ data:
         {{- if (eq .Values.mongo.auth.enabled true) }}
         username: {{ .Values.mongo.auth.username }}
         password: {{ .Values.mongo.auth.password }}
+        authSource: {{ .Values.mongo.auth.source | default "admin" }}
```
brasseld, line 54: Just wondering why we would want to move "admin" as default authSource ?
To be consistent with the code, I would prefer to keep `gravitee`.
See https://github.com/gravitee-io/gravitee-repository-mongodb/blob/master/src/main/java/io/gravitee/repository/mongodb/common/MongoFactory.java#L205

apim/3.x/templates/gateway/gateway-configmap.yaml
```diff
@@ -57,6 +57,7 @@ data:
         {{- if (eq .Values.mongo.auth.enabled true) }}
         username: {{ .Values.mongo.auth.username }}
         password: {{ .Values.mongo.auth.password }}
+        authSource: {{ .Values.mongo.auth.source | default "admin" }}
```
brasseld, line 60: Same as previous comment.

brasseld (changes requested):

apim/3.x/templates/api/api-configmap.yaml
Samaruq: I was confused for a second. the `values.yaml` has `admin` listed but the code uses `gravitee`. I deleted that commit now.

Samaruq: @brasseld , should I change the value of the auth source in the `values.yaml` to `gravitee` as well for consistency?

brasseld: ok, now I understand. So you're right, `admin` should be set by default, otherwise it will be no more compatible with previous chart versions.

Samaruq: I think `gravitee` is the right default though, because if you did not specify `.Values.mongo.auth.source` in your `values.yaml`, `gravitee` is the one that is used behind the scenes. In all other cases, what you specify is what is used.

brasseld: Indeed, good point!

So I think that's good.

Samaruq:
@brasseld, just noticed I forgot to include the auth source property in the `ratelimit` section of the `gateway0configmap` template. Will update my branch in a minute and submit PR again.

PR #89, opened 2021-02-23
[apim3] Enable or disable Elasticsearch ILM (Index Lifecycle Management)

ludo-berthelot:


apim/3.x/templates/api/api-configmap.yaml
```diff
@@ -129,6 +129,15 @@ data:
           username: {{ .Values.es.security.username }}
           password: {{ .Values.es.security.password }}
         {{- end }}
+        {{- if (eq .Values.es.lifecyle.enabled true) }}
+        lifecyle:
```
brasseld, line 133: I could be wrong, but I think ILM policies can only be configured from the gateway, and not from the management-api.

brasseld (changes requested):

apim/3.x/templates/api/api-configmap.yaml
ludo-berthelot: That's an error. I'm correcting.

ludo-berthelot: It's done

PR #95, opened 2021-04-01
am: update Chart missing values

tcompiegne:
fixes gravitee-io/issues#5312


leleueri (changes requested):
Two small comments before validate it

leleueri (approved):

PR #102, opened 2021-04-20
Add support for Hazelcast cache

boomaker:
Add support for new Hazelcast cache in Gravitee APIM 3.7
See
- https://www.gravitee.io/article/gravitee-io-api-platform-3-7
- https://github.com/gravitee-io/issues/issues/599

ehcache remains the default.

hazelcast.xml will be built based on default file I gathered from Gravitee APIM + `<network>` section is inspired from Gravitee Alert Engine + custom `<map>` blocks are possible using Values.cache.config
NB: I never used Hazelcache myself, if someone have knowledge, feel free to review my implementation 😊 Or other solution is to make hazelcast.xml totally customizable 🤔

### DIFF
--- a/apim/3.x/templates/gateway/gateway-configmap.yaml
+++ b/apim/3.x/templates/gateway/gateway-configmap.yaml
@@ -142,7 +142,7 @@ data:
         {{- end }}
       {{- end }}
     cache:
-      type: ehcache
+      type: {{ .Values.cache.type | default "ehcache" }}
       enabled: true
 
     # Sharding tags configuration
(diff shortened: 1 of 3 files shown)


brasseld:
Hi @boomaker 

Thanks for your contribution, can you explain me the purpose of this pull request, I'm not too sure it's relevant.

Looking forward hearing from you.

boomaker:
Hi @brasseld 

Actually the cache is statically set to ehcache (see https://github.com/gravitee-io/helm-charts/blob/e1b8da673fb63ed97f3ff5144e4a7b8c5f3afd60/apim/3.x/templates/gateway/gateway-configmap.yaml#L145)

Without the changes I suggest, it is impossible to use hazelcast to share the cache between gateway pods, for example API cache resources.

But feel free to correct me if I am wrong :)

brasseld:
As far as I know, this property is never read. 
We have added support for Hazelcast cache in 3.6, by using the cache resource plugin, but this later does not belong to this particular policy.

boomaker:
Ok, but anyway, we have to configure hazelcast.xml file to share cache resources between gateways, right?
As specified in https://docs.gravitee.io/apim/3.x/apim_installguide_cache.html#how_are_cache_resources_managed_in_apim

This PR also permits to configure this file.

PR #110, opened 2021-05-28
feat(cockpit): add oidc configuration

jgiovaresco:
Add the OIDC idp configuration.

I took the opportunity to extract the authentication config into `values.yaml`. Therefore we won't need to update the charts to update the configuration for `github`, `google` and `oidc`.
However, we would need to update the chart if we add a new IDP.

### DIFF
--- a/cockpit/templates/api/api-configmap.yaml
+++ b/cockpit/templates/api/api-configmap.yaml
@@ -191,47 +191,28 @@ data:
     auth:
       callbackUrl: {{ .Values.api.ingress.path }}/auth/login/callback
 
-      {{- if and .Values.authentication.github .Values.authentication.github.clientId }}
+      {{- if .Values.authentication.github }}
       github:
-        userAutorizationUri: https://github.com/login/oauth/authorize
-        accessTokenUri: https://github.com/login/oauth/access_token
-        userProfileUri: https://api.github.com/user
-        codeParameter: code
-        responseType: code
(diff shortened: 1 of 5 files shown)


cockpit/templates/api/api-configmap.yaml
```diff
@@ -191,31 +191,24 @@ data:
...
-        clientId: {{ .Values.authentication.google.clientId }}
-        clientSecret: {{ .Values.authentication.google.clientSecret }}
         redirectUri: https://{{ index .Values.api.ingress.hosts 0 }}{{ .Values.api.ingress.path }}/auth/login/callback?provider=google
-        scopes: openid,profile,email
+{{ toYaml .Values.authentication.google | indent 8 }}
       {{- end }}
+
+      {{- if .Values.authentication.oidc }}
+      oidc:
+        redirectUri: https://{{ index .Values.api.ingress.hosts 0 }}{{ .Values.api.ingress.path }}/auth/login/callback?provider=oidc
```
brasseld, line 208: IMHO, it's better to allow the user to override the `redirectUri` by himself.
So I would propose to check if `.Values.authentication.oidc.redirectUri` is defined. If it is, use it, otherwise, calculate it.

brasseld (changes requested):

brasseld:
@jgiovaresco also, could you update changelog and release version please.

PR #118, opened 2021-06-25
feat: remove default admin (in memory) user for more security

bmixte:
Hello,

We want to manage global algorithm encryption in values and remove the default admin user.

Breaking change : no more default admin in memory user.
If helm chart use want to keep the default admin user they must add the following entry in the values :

```yaml
    extraInMemoryUsers: |
      - user:
        username: <admin (or another username if you want to secure your access)>
        password: <your secudred password (bcrypt encoded if you let the default algo)>
        roles: ORGANIZATION:ADMIN, ENVIRONMENT:ADMIN
```

K.R.
Benjamin.

### DIFF
--- a/apim/3.x/templates/api/api-configmap.yaml
+++ b/apim/3.x/templates/api/api-configmap.yaml
@@ -163,16 +163,10 @@ data:
           # - bcrypt : passwords are hashed with bcrypt
           # - none : passwords are not hashed/encrypted
           # default value is bcrypt
-          password-encoding-algo: bcrypt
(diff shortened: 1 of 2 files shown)


brasseld:
Hey @bmixte 

Thanks for this PR.

I'm wondering if it's not better / easier if we are adding a boolean to enable / disable admin account like 'adminAccountEnable'.
That way, there will be no breaking change at all since the default value for this flag would be for the admin account to be enabled.

WDYT?

bmixte:
Ok, for me, i will make the change on chart and update this P.R.

brasseld:
Thanks a lot @bmixte 👍

PR #120, opened 2021-06-28
fix(designer): Remove useless API ingress path regex capture

brasseld:
Closes gravitee-io/issues#5797


Spitfire1900:
This may break with the Helm chart's default annotation:
```
annotations:
    nginx.ingress.kubernetes.io/rewrite-target: /$2
```

See https://helm.gravitee.io/helm/designer/designer-1.0.6.tgz, resource designer-1.0.6.tar\designer\values.yaml, line 115.

PR #129, opened 2021-06-30
chore: chore: Improve Cockpit deployment configurability

brasseld:
Closes gravitee-io/gravitee-cockpit#763

### DIFF
--- a/am/values.yaml
+++ b/am/values.yaml
@@ -173,6 +173,15 @@ api:
   # https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass
   priorityClassName: ""
 
+  env: []
+  #   - name: ENV_VARIABLE
+  #     value: ENV_VARIABLE_VALUE
+  #   - name: ENV_VARIABLE_WITH_FROM
+  #     valueFrom:
+  #       configMapKeyRef:
+  #         name: special-config
+  #         key: SPECIAL_LEVEL
+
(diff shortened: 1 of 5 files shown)


am/values.yaml
```diff
@@ -173,6 +173,15 @@ api:
   # https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass
   priorityClassName: ""
 
+  env: []
```
jgiovaresco, line 176: Did you see that is an AM file, should we include it in the PR? 🤔

cockpit/values.yaml
```diff
@@ -127,14 +127,60 @@ api:
...
+    tcpSocket:
+      port: http
+    initialDelaySeconds: 30
+    periodSeconds: 30
+    failureThreshold: 3
+
+  podAnnotations: {}
+    # iam.amazonaws.com/role: es-cluster
+
+  # How long to wait for APIM Management API pods to stop gracefully
```
jgiovaresco, line 168:
```suggestion
  # How long to wait for Cockpit Management API pods to stop gracefully
```

jgiovaresco (approved):

am/values.yaml
brasseld: Yep, sorry for that. I've missed this from my previous AM PR :(

brasseld:
FP

PR #140, opened 2021-08-12
Fix typos in ES reporter lifecycle definition

boomaker:
ILM for Elastic Search indexes templates does not work actually because of theses typos.


brasseld:
thanks for the contribution @boomaker

PR #141, opened 2021-08-18
Update gravitee-am ingress descriptor to support k8s v1.22+

famaridon:
This pull request add compatibility with k8s 1.22+ whose remove support of **extensions/v1beta1** and **networking.k8s.io/v1beta1** apiVersion for Ingress resource kind
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122

And this remove warning when deploy on k8s v1.19+.

```

This PR is based on the default Helm chart Ingress template. I just adapt it to keep the same configuration in `values.yaml`.

### DIFF
--- a/am/templates/api/api-ingress.yaml
+++ b/am/templates/api/api-ingress.yaml
@@ -3,7 +3,14 @@
 {{- $serviceAPIName := include "gravitee.api.fullname" . -}}
 {{- $serviceAPIPort := .Values.api.service.externalPort -}}
 {{- $ingressPath   := .Values.api.ingress.path -}}
-apiVersion: {{ template "ingress.apiVersion" . }}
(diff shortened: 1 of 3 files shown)


brasseld:
Hi @famaridon 

Thanks for the proposal.

Instead of duplicating the same piece of code to discover the ingress version, can you just update this part: https://github.com/gravitee-io/helm-charts/blob/master/am/templates/_helpers.tpl#L58

This will allow to centralize the different tests for each of the Ingress.

Thanks,

famaridon:
Hi @brasseld.

Thanks for the recommendations.

I improved the code based on it. But I'm currently not able to test with kubernates version < 1.20 (I updated all my clusters).

olevitt:
Hi, 
Kubernetes 1.22 is now out for 2 months. We are in the process of upgrading our clusters and would really benefit from this being merged.  
LGTM :+1:

PR #148, opened 2021-09-04
Update Helm Chart w/ PostgresSQL info

tiffanymarchi:
Updated ReadMe partial to include details on installing PostgresSQL in addition to MongoDB. @marcambier Let's sync on this and we can go over all these changes.

### DIFF
--- a/apim/3.x/README.adoc
+++ b/apim/3.x/README.adoc
@@ -1,78 +1,97 @@
 [[graviteeio-api-management-helm-chart]]
 == Gravitee.io API Management Helm Chart
 
-*Chart supported versions: 3.0.x and higher*
+*Chart supported versions: 3.0.x and higher*. 
 
 === Components
 
 This chart will deploy the following:
 
-* Gravitee Management API
-* Gravitee Management UI
-* Gravitee Portal UI
-* Gravitee Gateway
-* MongoDB replica-set (optional dependency)
-* Elasticsearch cluster (optional dependency)
+* Gravitee Management API.
+* Gravitee Management UI.
+* Gravitee Portal UI.
+* Gravitee Gateway.
+* MongoDB replica set  or PostgresSQL (optional dependency).
+* Elasticsearch Cluster (optional dependency).
 
-=== Installing
+=== 1. Install The Gravitee Helm Chart Repo
 
-* Add the Gravitee.io helm charts repo
+* Add the official Gravitee.io Helm chart repo: 
 +
 ....
 $ helm repo add graviteeio https://helm.gravitee.io
 ....
-* Install it
-+
-....
-$ helm install --name graviteeio-apim3x graviteeio/apim3
-....
 
-=== Create a chart archive
 
-To package this chart directory into a chart archive, run:
+=== 2. Create The Configuration File & Set Custom Values 
+Before you continue installing the Helm chart, create a `values.yml` configuration file and set any applicable custom values in it. 
 
-....
-$ helm package .
-....
 
-=== Installing the Chart
-
-To install the chart from the Helm repository with the release name
-`+graviteeio-apim3x+`:
-
-[source,bash]
 ----
-$ helm install --name graviteeio-apim3x graviteeio/apim3
+# Management console version
+ui:
+  image:
+    tag: 3.10.0
+
+# Gateway version
+gateway:
+  image:
+    tag: 3.10.0
+    
+# Portal version
+portal:
+  image:
+    tag: 3.10.0
+
+# Management API version
+api:
+  image:
+    tag: 3.10.0
+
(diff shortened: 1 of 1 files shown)


apim/3.x/README.adoc
```diff
@@ -208,7 +263,39 @@ pfx) |`+null+`
...
+To install the chart from the Helm repository with the release name
+`+graviteeio-apim3x+`, run:
+
+[source,bash]
+----
+$ helm install --name graviteeio-apim3x graviteeio/apim3
+----
+
+
+*Note:* If you're using Helm 3, this name parameter is no longer valid. Instead, please
```
marcambier, line 280: Helm v3 is the current version (has been released on 2019).
So I think default commands should be compatible with helm v3 (without --name parameters).
And maybe add a compatibility note for older versions (or just add helm V3 as a prerequisite, at start of document)

apim/3.x/README.adoc
```diff
@@ -150,18 +186,37 @@ define the following configuration options:
...
+
+
+==== PostgreSQL Database (via JDBC Connection) 
+To install a new PostgresSQL database, use the command below and update the `username`, `password`, and `databasename` parameters: 
+----
+helm install --set postgresqlUsername=postgres --set postgresqlPassword=P@ssw0rd 
+--set postgresqlDatabase=graviteeapim postgres-apim bitnami/postgresql
+----
+
+NOTE: When you install APIM, it automatically uses the default value for the host and runs on apim.example.com. To change this default value to a specific host name for your APIM installation, you must modify the `gateway.ingress.hosts` parameter (listed in the link:/apim/3.x/apim_installguide_kubernetes.html#gravitee_gateway[Gravitee Gateway table] below) in the `values.yml` configuration file.
```
marcambier, line 210: If users want to use a custom hostname, they have to change multiple values in their configuration file, not only one : there is a discinct hostname configuration for each app (gateway/api/console/portal).
For example, you can search in the documentation all the values which have `[apim.example.com]` as default.

apim/3.x/README.adoc
```diff
@@ -208,7 +263,39 @@ pfx) |`+null+`
...
+
+
+=== 4. Install The Gravitee Helm Chart
+
+To install the chart from the Helm repository with the release name
+`+graviteeio-apim3x+`, run:
+
+[source,bash]
+----
+$ helm install --name graviteeio-apim3x graviteeio/apim3
```
marcambier, line 276: This command installs APIM.
If we want the configuration file created previously to be considered, we have to add the `-f values.yaml` parameter

apim/3.x/README.adoc
```diff
@@ -1,78 +1,97 @@
...
+For PostgrestSQL, use the information below in `values.yml` and replace the `username`, `password`, 
+`URL` and `database name` with details for your specific instance. 
 
-....
-$ helm install apim3-3.0.0.tgz
-....
-
-=== Configuration
+----
+jdbc:
```
marcambier, line 71: I think it could be confusing that the postgresql part of the configuration file is located here,
But the mongo part of the configuration file is located under "Configure Optional Dependencies" chapter

apim/3.x/README.adoc
```diff
@@ -150,18 +186,37 @@ define the following configuration options:
...
 See
 https://github.com/helm/charts/tree/master/stable/mongodb-replicaset[MongoDB
-replicaset] for detailed documentation on helm chart.
+replica set] for detailed documentation on Helm chart.
+
+
+==== PostgreSQL Database (via JDBC Connection) 
+To install a new PostgresSQL database, use the command below and update the `username`, `password`, and `databasename` parameters: 
+----
+helm install --set postgresqlUsername=postgres --set postgresqlPassword=P@ssw0rd
```
marcambier, line 206: Here, we provide the helm command to install postgres.
But in the mongo part, we do not provide the helm command to install mongo.
Which would be, for example `helm install mongodb bitnami/mongodb --set auth.rootPassword=r00t`

apim/3.x/README.adoc
```diff
@@ -82,13 +101,30 @@ chart)
 |`+ldap.enabled+` |Enable LDAP login |false
 |===
 
-==== Mongo
 
-===== MongoDB connections
+
+=== 3. Configure  Optional Dependencies 
+
+Specify each parameter using the `+--set key=value[,key=value]+`
```
marcambier, line 108: Suggestion, maybe those lines from 108-120 are like a "helm install command tip", and should be located in a dedicated section ?
It may be confusing cause it seems user has to launch this `$ helm install --name my-release -f values.yaml gravitee` command here. But it's just here as an example ?

tiffanymarchi:
Closing PR (see #149 Update ReadMe Include instead).

PR #149, opened 2021-09-06
Update ReadMe Include

tiffanymarchi:
Rewrite of info to install PostgresSQL via JDBC. @marcambier, could you please take a look at this?

### DIFF
--- a/apim/3.x/README.adoc
+++ b/apim/3.x/README.adoc
@@ -1,33 +1,33 @@
 [[graviteeio-api-management-helm-chart]]
+[[graviteeio-api-management-helm-chart]]
 == Gravitee.io API Management Helm Chart
 
-*Chart supported versions: 3.0.x and higher*
+Installing APIM on Kubernetes is easy with the help of our Helm Chart. **This Helm Chart supports versions: 3.0.x and higher** and will deploy the following: 
 
-=== Components
+* Gravitee Management API.
+* Gravitee Management UI.
+* Gravitee Portal UI.
+* Gravitee Gateway.
+* MongoDB replica set or PostgresSQL (optional dependency).
+* Elasticsearch Cluster (optional dependency).
 
-This chart will deploy the following:
+=== Prerequisites
 
-* Gravitee Management API
-* Gravitee Management UI
-* Gravitee Portal UI
-* Gravitee Gateway
-* MongoDB replica-set (optional dependency)
-* Elasticsearch cluster (optional dependency)
+. https://helm.sh/docs/intro/install/[Helm v3].
+. MongoDB or PostgreSQL. (Not to worry, this Helm Chart will deploy these.)
+. Elasticsearch. (This Helm Chart will also deploy this.)
 
-=== Installing
 
-* Add the Gravitee.io helm charts repo
-+
+
+=== Install The Helm Chart Repo
+
+Add the Gravitee.io Helm charts repo using the commmand below: 
+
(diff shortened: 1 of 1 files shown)


tiffanymarchi:
@marcambier -- Should I add this info? 

----
# Management console version
ui:
  image:
    tag: 3.11.0

# Gateway version
gateway:
  image:
    tag: 3.11.0
    
# Portal version
portal:
  image:
    tag: 3.11.0

# Management API version
api:
  image:
    tag: 3.11.0

# gateway proxy configuration
# (remove '/gateway' prefix from URL before sending request)
gateway:
  ingress:
    path: /gateway/(.*)
    annotations:
      nginx.ingress.kubernetes.io/rewrite-target: /$1

# ES endpoint
es:
  endpoints:
    - http://elastic-elasticsearch-data.default.svc.cluster.local:9200
----

apim/3.x/README.adoc
```diff
@@ -37,42 +37,62 @@ $ helm package .
...
-To install the chart from the Helm repository with the release name
-`+graviteeio-apim3x+`:
+
+Now, install the chart from the Helm repo with the release name
+
+`+graviteeio-apim3x+`, run:
 
 [source,bash]
 ----
 $ helm install --name graviteeio-apim3x graviteeio/apim3
```
marcambier, line 47: This is a helm v2 command, the helm v3 is without the --name parameter : 
`helm install graviteeio-apim3x graviteeio/apim3`

apim/3.x/README.adoc
```diff
@@ -1,33 +1,33 @@
...
 $ helm repo add graviteeio https://helm.gravitee.io
 ....
-* Install it
-+
-....
-$ helm install --name graviteeio-apim3x graviteeio/apim3
-....
 
-=== Create a chart archive
+=== Create A Chart Archive
```
marcambier, line 30: I think creating this archive is not a part of the installation process.
Anyway, this command will do nothing good cause it uses the current user directory ; and we don't indicate where to launch it.
Maybe it's just confusing and we can remove this "Create A Chart Archive" section ?

apim/3.x/README.adoc
```diff
@@ -163,6 +187,38 @@ See
...
 replicaset] for detailed documentation on helm chart.
 
+==== PostgresSQL (via JDBC Connection) 
+To install a new PostgresSQL database, use the command below and update the `username`, `password`, and `databasename` parameters: 
+----
+helm install --set postgresqlUsername=postgres --set postgresqlPassword=P@ssw0rd 
+--set postgresqlDatabase=graviteeapim postgres-apim bitnami/postgresql
+----
+
+NOTE: When you install APIM, it automatically uses the default value for the host and runs on `apim.example.com`. To use a specific host name, search through our documentation and replace any values in the `values.yml` configuration file that use the default value of `apim.example.com`.
```
marcambier, line 197: Nice note :+1: 
But I don't think it's related to PosgtreSQL configuration.
Maybe we can move it in another section ? For example the "Helm Install Command Tips" ?

marcambier:
> @marcambier -- Should I add this info?

This is a sample configuration file I used to run APIM locally.
All parameters it contains are listed and explained in the "configuration" section of the documentation.

So I don't think it's needed to add this example ;
Users installing APIM on Kubernetes will have to create their own configuration file according to their environment ; Relying on the parameters list we provide in this documentation

PR #152, opened 2021-09-15
fix: trouble with automatic management URL

bmixte:
Our management public URL is not the ingress one.
We have a WAF Service in front of our ingress controller.
We change values to be able to change from default ingress value.

### DIFF
--- a/apim/3.x/templates/api/api-configmap.yaml
+++ b/apim/3.x/templates/api/api-configmap.yaml
@@ -34,8 +34,12 @@ data:
     {{- end }}
 
     management:
-      {{- if .Values.ui.ingress.enabled }}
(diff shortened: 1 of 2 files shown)


apim/3.x/templates/api/api-configmap.yaml
```diff
@@ -34,8 +34,12 @@ data:
     {{- end }}
 
     management:
-      {{- if .Values.ui.ingress.enabled }}
-      url: https://{{ index .Values.ui.ingress.hosts 0 }}{{ regexFind "/[a-zA-Z]+" .Values.ui.ingress.path }}/
+      {{- if .Values.management.url }}
+      url: {{ Values.management.url}}
+      {{- else }}
```
brasseld, line 39: else if ?

PR #157, opened 2021-09-16
[feature] add harbor dockerhub_cache / private repo

dstdvb:
Some kubernetes environments have no direct internet connection and
rely on a solution like harbor to proxy the container downloads.

There was no clean solution to configure this, as the image paths for
the alpine init containers where hardcoded in the template.

This change makes use of a harbor repository easily and configurable.
But it can also be used for other container repo's or caches.

In our situation harbor's dockerhub-cache runs at:
`some.harbor.local/dockerhub-cache/<repo>/<image>`

configure in values.yml:
```
harbor:
  enabled: true
  repository:
    host: some.harbor.local
    path: dockerhub-cache
    official: library
```

resulting alpine image: `some.harbor.local/dockerhub-cache/library/alpine`
resulting gravitee image: `some.harbor.local/dockerhub-cache/graviteeio/apim-<image>:<tag>`

### DIFF
--- a/apim/3.x/templates/api/api-deployment.yaml
+++ b/apim/3.x/templates/api/api-deployment.yaml
@@ -86,7 +86,11 @@ spec:
       initContainers:
         {{- if and .Values.jdbc.driver (eq .Values.management.type "jdbc") }}
         - name: get-repository-jdbc-ext
+          {{- if (eq .Values.harbor.enabled true) }}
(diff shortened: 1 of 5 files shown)


brasseld:
Thanks for the contribution @dstdvb .

As we didn't want to add support for harbor specifically, we build an other way to manage initContainers image / tag.

Closed by https://github.com/gravitee-io/helm-charts/commit/d53ffeabea2ad527b775fd69f878575acf1c9de5

PR #185, opened 2021-11-25
fix: nginx must pass if-match header

ytvnr:
gravitee-io/issues#6624

## Context

In APIM, if an API is modified by multiple users at the same time, only the first can modify it. The following wil need to refresh the page. This mechanism is implemented by using ETag and if-match headers;

## Resolution

In our configuration, if-match header is overridden with an empty string. This fix tells to nginx to pass the header as usual.

⚠️ unable to test locally

### DIFF
--- a/apim/3.x/CHANGELOG.md
+++ b/apim/3.x/CHANGELOG.md
@@ -5,6 +5,7 @@ This file documents all notable changes to [Gravitee.io API Management 3.x](http
 ### 3.1.31
 
 - [X] Add support for ILM managed indexes
+- [X] Remove the empty override of 'if-match' header
 
 ### 3.1.30 
 
--- a/apim/3.x/values.yaml
+++ b/apim/3.x/values.yaml
@@ -528,7 +528,7 @@ api:
         - apim.example.com
       annotations:
         kubernetes.io/ingress.class: nginx
-        nginx.ingress.kubernetes.io/configuration-snippet: "etag on;\nproxy_pass_header ETag;\nproxy_set_header if-match \"\";\n"
+        nginx.ingress.kubernetes.io/configuration-snippet: "etag on;\nproxy_pass_header ETag;\nproxy_pass_header if-match;\n"
         # kubernetes.io/tls-acme: "true"
       #tls:
         # Secrets must be manually created in the namespace.


apim/3.x/values.yaml
```diff
@@ -547,7 +547,7 @@ api:
         - apim.example.com
       annotations:
         kubernetes.io/ingress.class: nginx
-        nginx.ingress.kubernetes.io/configuration-snippet: "etag on;\nproxy_pass_header ETag;\nproxy_set_header if-match \"\";\n"
+        nginx.ingress.kubernetes.io/configuration-snippet: "etag on;\nproxy_pass_header ETag;\nproxy_pass_header if-match;\n"
```
brasseld, line 550: @ytvnr do you really we need if for portal API ?

apim/3.x/values.yaml
```diff
@@ -764,7 +764,7 @@ gateway:
       #     nginx.ingress.kubernetes.io/enable-rewrite-log: "true"
       #     kubernetes.io/app-root: /gateway
       #     kubernetes.io/rewrite-target: /gateway
-      #     nginx.ingress.kubernetes.io/configuration-snippet: "etag on;\nproxy_pass_header ETag;\nproxy_set_header if-match \"\";\n"
+      #     nginx.ingress.kubernetes.io/configuration-snippet: "etag on;\nproxy_pass_header ETag;\nproxy_pass_header if-match;\n"
```
brasseld, line 767: @ytvnr same here, I think it's useless here.

apim/3.x/values.yaml
```diff
@@ -846,7 +846,7 @@ gateway:
     annotations:
       kubernetes.io/ingress.class: nginx
       nginx.ingress.kubernetes.io/ssl-redirect: "false"
-      # nginx.ingress.kubernetes.io/configuration-snippet: "etag on;\nproxy_pass_header ETag;\nproxy_set_header if-match \"\";\n"
+      # nginx.ingress.kubernetes.io/configuration-snippet: "etag on;\nproxy_pass_header ETag;\nproxy_pass_header if-match;\n"
```
brasseld, line 849: @ytvnr same here, I think it's useless here.

apim/3.x/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 This file documents all notable changes to [Gravitee.io API Management 3.x](https://github.com/gravitee-io/helm-charts/tree/master/apim/3.x) Helm Chart. The release numbering uses [semantic versioning](http://semver.org).
 
+### 3.1.32
+
+- [X] Remove the empty override of 'if-match' header
```
brasseld, line 7: Move back to 3.1.31

apim/3.x/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 name: apim3
 # When the version is modified, make sure the artifacthub.io/changes list is updated
 # Also update CHANGELOG.md
-version: 3.1.31
+version: 3.1.32
```
brasseld, line 5: Move back to 3.1.31

brasseld (changes requested):
Request changes as per comments.
Also, please update changelog the Chart.yaml.

apim/3.x/values.yaml
ytvnr: since we can configure cors for portal and set allowed header list I guess it's preferable to pass the original header.
but I maybe wrong
wdyt ?

https://github.com/gravitee-io/gravitee-api-management/blob/724f821317198ad36f151be3e4caeaafd6fdca6d/gravitee-apim-rest-api/gravitee-apim-rest-api-portal/gravitee-apim-rest-api-portal-security/src/main/java/io/gravitee/rest/api/portal/security/config/GraviteeCorsConfiguration.java#L37

apim/3.x/values.yaml
ytvnr: done

apim/3.x/values.yaml
ytvnr: done

apim/3.x/CHANGELOG.md
ytvnr: done

apim/3.x/Chart.yaml
ytvnr: done

apim/3.x/values.yaml
brasseld: Yes, but usage of if-match is not relative to the cors itself. it is only used for checking that there is no concurrent changes to an API based on ETag. So I would say that this one is useless.

ytvnr: understood! 
rolled back :)

PR #208, opened 2022-01-25
Update README.adoc

tomgeudens:
ASCIIDOC does not like explicit list numbering. Fixed.


marcambier (approved):
OK to me.
But maybe we would better use a conventional commit message, starting with "docs:" ?
https://www.conventionalcommits.org/

PR #219, opened 2022-02-24
Handle jwt attributes in APIM Management API

gaetanmaisse:
**Issue**

https://github.com/gravitee-io/issues/issues/7023

**Description**

I added `jwt` attributes in APIM Management API configmap.
Regarding the naming I followed the convention started by `cookie-domain: {{ .Values.jwt.cookie.domain }}`

### DIFF
--- a/apim/3.x/templates/api/api-configmap.yaml
+++ b/apim/3.x/templates/api/api-configmap.yaml
@@ -342,13 +342,13 @@ data:
       secret: {{ .Values.jwtSecret }}
       {{- if .Values.jwt }}
       # Allows to define the end of validity of the token in seconds (default 604800 = a week)
-      #expire-after: 604800
+      expire-after: {{ .Values.jwt.expireAfter | default 604800 }}
       # Allows to define the end of validity of the token in seconds for email registration (default 86400 = a day)
-      #email-registration-expire-after: 86400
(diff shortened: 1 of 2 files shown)


phiz71:
Shouldn't we use this file to store default values ?
https://github.com/gravitee-io/helm-charts/blob/master/apim/3.x/values.yaml

phiz71:
Is it worth adding a unit test ?

gaetanmaisse:
>Shouldn't we use this file to store default values?
[https://github.com/gravitee-io/helm-charts/blob/master/apim/3.x/values.yaml](https://github.com/gravitee-io/helm-charts/blob/master/apim/3.x/values.yaml?rgh-link-date=2022-02-24T21%3A21%3A33Z)

I don't know I did like some of the other attributes and as it's done on AM side 🤷🏻 

--

> Is it worth adding a unit test?

I wasn't able to run them locally yesterday, looks better now... well, they are failing, not a surprise as there is no CI 😞 
So I will start by creating a PR to run tests in CI and fix the failing ones

gaetanmaisse:
Looks like you faced the same issue @phiz71 I will update https://github.com/gravitee-io/helm-charts/pull/191 then ;)

phiz71 (approved):

PR #220, opened 2022-02-25
Provide URLs for console API and UI

Andy2003:
**Issue**

https://github.com/gravitee-io/issues/issues/6224

**Description**

Setup console URLs for cloud setup

**Additional context**

- https://github.com/gravitee-io/gravitee-api-management/pull/1318
- https://github.com/gravitee-io/gravitee-api-management/pull/1318#discussion_r814631209

### DIFF
--- a/apim/3.x/templates/api/api-configmap.yaml
+++ b/apim/3.x/templates/api/api-configmap.yaml
@@ -37,7 +37,7 @@ data:
       {{- if .Values.management.url }}
       url: {{ .Values.management.url}}
       {{- else if .Values.ui.ingress.enabled }}
-      url: https://{{ index .Values.ui.ingress.hosts 0 }}{{ .Values.ui.ingress.basePath }}/
+      url: https://{{ index .Values.ui.ingress.hosts 0 }}{{ regexFind "/[a-zA-Z]+" .Values.ui.ingress.path }}/
       {{- end }}
       type: {{ .Values.management.type | default "mongodb" }}
       {{- if or (eq .Values.management.type "mongodb") (kindIs "invalid" .Values.management.type) }}
@@ -79,8 +79,7 @@ data:
       api:
         url: "https://{{index .Values.api.ingress.management.hosts 0 }}{{ .Values.api.ingress.management.path }}"
       ui:
-        url: "https://{{index .Values.ui.ingress.hosts 0 }}{{ .Values.ui.ingress.basePath }}"
-
+        url: "https://{{index .Values.ui.ingress.hosts 0 }}{{ regexFind "/[a-zA-Z]+" .Values.ui.ingress.path }}"
     services:
       core:
         http:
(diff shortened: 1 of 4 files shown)


phiz71:
We try to follow [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/), could you reword your commit like this: `feat(apim): provide...` ? Thx 🙏🏻

phiz71 (approved):

apim/3.x/templates/api/api-configmap.yaml
```diff
@@ -75,6 +75,12 @@ data:
         {{- end }}
       {{- end }}
 
+    console:
+      api:
+        url: "https://{{index .Values.api.ingress.management.hosts 0 }}{{ .Values.api.ingress.management.path }}"
+      ui:
+        url: "https://{{index .Values.ui.ingress.hosts 0 }}{{ .Values.ui.ingress.path }}"
```
Andy2003, line 82: this results in 

```yaml
    console:
      ui:
        url: "https://api-portal.gravitee.local/console(/|$)(.*)"
```

for my local test setup, chis is not a valid URL.

@phiz71 iny suggestion?

Andy2003: resolved issue via regex

phiz71: Ok, I checked with @brasseld.
Could you please:
 - remove `(/|$)(.*)` from the values.yml file.
 ```
 path: /console(/|$)(.*) ==>  path: /console
 ```
 - add `(/|$)(.*)` in ui-ingress.yaml file
 ```
path: {{ $ingressPath }} ==>  path: {{ $ingressPath }}(/|$)(.*)
 ```
 ?

Then you will be able to use `{{ .Values.ui.ingress.path }}` without regex filter

apim/3.x/values.yaml
```diff
@@ -1168,7 +1168,9 @@ ui:
     runAsNonRoot: true
   ingress:
     enabled: true
-    path: /console(/|$)(.*)
+    basePath: /console
```
phiz71, line 1171: changing path to basePath will be a breaking change. If some users already have a path defined, it will no longer be taken into account.

phiz71:
I'm also thinking that if some users have configured `path: /mypath/foo(/|$)(.*)` in their values, we should not append the regex 🙁

Andy2003:
> I'm also thinking that if some users have configured `path: /mypath/foo(/|$)(.*)` in their values, we should not append the regex

this is exactly the reason I added `basePath`.

So if we want to be backwards compatible we need to check for the existence of `path` and do the regex magic,  when it exists.

With the given configuration `path: /mypath/foo(/|$)(.*)`, also the approach mentioned in https://github.com/gravitee-io/helm-charts/pull/220#discussion_r814724186 will break.

gaetanmaisse:
FYI a meeting will be planned this week with other helm-interested-fox at Gravitee to discuss this PR, I will keep you updated

PR #222, opened 2022-03-07
feat(am): Introduce values to enable and configure notification about…

leleueri:
… domain certificates expiry

related gravitee-io/issues#6833


am/templates/api/api-configmap.yaml
```diff
@@ -144,6 +144,50 @@ data:
...
+          {{- if .Values.api.services.certificate.expiryEmailSubject }}
+        expiryEmailSubject: {{ .Values.api.services.certificate.expiryEmailSubject | quote }}
+          {{- end }}
+        {{- end }}
+    {{- end }}
+  {{- end }}
+
+    notifiers:
+      email: 
+        enabled: {{ .Values.api.notifiers.email.enabled }}
```
brasseld, line 167: @leleueri What is the reason of not reusing the smtp.* values and use dedicated one? Do we think that they can (really) use an other SMTP for the two purposes?

leleueri: @brasseld not it will be the same value for sure but I did it to be align with the APIM way to configure notifiers.

brasseld: Yes, I was assuming that it was for aligning it with APIM.... but... it's currently a pain to manage SMTP on APIM side because of that.
I would consider to provide ability to override the smtp for the notifiers, but if there is no override, reuse the one defined for SMTP.
wdyt ?

leleueri: :+1:

brasseld: Indeed, I would be interested to see how you leverage that... Facing same issue with other products ;)

leleueri: I use notifier values in priority and fallback to smtp settings if missing.
For truststore, this is a dedicated entry into the notifier section.

farmborough (approved):
👍

PR #228, opened 2022-03-17
APIM: Open port for core services of the management API

gaetanmaisse:
**Issue**

https://github.com/gravitee-io/issues/issues/6895

**Description**

Open service core port for management API as it was already done for the gateway.

### DIFF
--- a/apim/3.x/templates/api/api-deployment.yaml
+++ b/apim/3.x/templates/api/api-deployment.yaml
@@ -107,6 +107,10 @@ spec:
           ports:
             - name: {{ .Values.api.service.internalPortName }}
               containerPort: {{ .Values.api.service.internalPort }}
+            {{- if .Values.api.http.services.core.http.enabled }}
+            - name: {{ .Values.api.name }}-hook
+              containerPort: {{ .Values.api.http.services.core.http.port | default "18083" }}
+            {{- end }}
           env:
             {{- if $plugins }}
             - name: GRAVITEE_PLUGINS_PATH_0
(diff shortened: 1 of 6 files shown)


brasseld:
Can you please add a test for this ?

gaetanmaisse:
@brasseld Done ✅

apim/3.x/templates/api/api-deployment.yaml
```diff
@@ -107,6 +107,10 @@ spec:
           ports:
             - name: {{ .Values.api.service.internalPortName }}
               containerPort: {{ .Values.api.service.internalPort }}
+            {{- if .Values.api.http.services.core.http.enabled }}
+            - name: {{ .Values.api.name }}-hook
```
brasseld, line 111: Not sure what's the meaning of `-hook` here. Shouldn't it be something like `-tech-api` ?

brasseld (changes requested):
See comments

apim/3.x/templates/api/api-deployment.yaml
gaetanmaisse: To be consistent with https://github.com/gravitee-io/helm-charts/blob/70d07a94244a55a79879cba445100b610cf977e4/apim/3.x/templates/gateway/gateway-deployment.yaml#L128

brasseld: Ok, I think that doesn't make sense. This one has been added in the past to the gateway for an other purpose (kubernetes hook), and I think it's been done wrongly at that time.

brasseld:
LGTM, did you try to deploy it?

PR #229, opened 2022-03-17
APIM: Add values to override the home folder of the APIM installation 

gaetanmaisse:
**Issue**

https://github.com/gravitee-io/issues/issues/6991

**Description**

Add values to override the home folder of the APIM installation


brasseld:
Can we do the same for all products to ensure consistency across products ?

guillaumelamirand:
Could you provide some validation on top of graviteeRoot and homePath ? for example by using a function in helper to build the path and adding possible missing /.

brasseld (changes requested):
Request changes as per previous comments

PR #230, opened 2022-03-25
fix: gateway cookie configuration

farmborough:
fixes gravitee-io/issues#7171

**Issue**

https://github.com/gravitee-io/issues/issues/7171

**Description**

Added cookie configuration to the gateway configmap.

**Additional context**

### DIFF
--- a/am/templates/gateway/gateway-configmap.yaml
+++ b/am/templates/gateway/gateway-configmap.yaml
@@ -15,6 +15,16 @@ data:
     # Gateway HTTP server
     http:
       port: {{ .Values.gateway.service.internalPort }}
+    {{- if .Values.gateway.http.cookie }}
+      cookie:
+        secure: {{ .Values.gateway.http.cookie.secure | default false }}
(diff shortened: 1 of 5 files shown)


am/templates/gateway/gateway-configmap.yaml
```diff
@@ -15,6 +15,12 @@ data:
     # Gateway HTTP server
     http:
       port: {{ .Values.gateway.service.internalPort }}
+      cookie:
```
leleueri, line 18: All the cookie block should be wrapped by a condition other wise we may face to some issue during the generation if as there are no default cookie settings into the values.yaml.

```
{{- if .Values.gateway.http.cookie }}
  cookie:
    secure: ...
{{- end }}
```

leleueri (changes requested):
In addition of the comment here after, can you update the changelog for the version 1.0.36 to inform uses that the sookie settings are now managed ?

Files to update:

* https://github.com/gravitee-io/helm-charts/blob/master/am/Chart.yaml#L26 
* https://github.com/gravitee-io/helm-charts/blob/master/am/CHANGELOG.md?plain=1#L7

PR #236, opened 2022-04-08
fix: add service accounts based on apim.managedServiceAccount flag in…

kamiiiel:
https://github.com/gravitee-io/_support_team/issues/142

### DIFF
--- a/cockpit/CHANGELOG.md
+++ b/cockpit/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 This file documents all notable changes to [Gravitee.io Cockpit](https://github.com/gravitee-io/helm-charts/tree/master/cockpit) Helm Chart. The release numbering uses [semantic versioning](http://semver.org).
 
+### 1.6.12
+
+- Fix ignoring the apim.managedServiceAccount in deployment files
+
 ### 1.6.11
 
 - fix API_URL environment variable name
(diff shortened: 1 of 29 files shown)


gaetanmaisse (changes requested):
A few feedbacks:
 - You also need to update the version here https://github.com/gravitee-io/helm-charts/blob/master/apim/3.x/Chart.yaml#L5
 - I think we should apply the same on AM and AE as there is the same bug

cockpit/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 This file documents all notable changes to [Gravitee.io Cockpit](https://github.com/gravitee-io/helm-charts/tree/master/cockpit) Helm Chart. The release numbering uses [semantic versioning](http://semver.org).
 
+### 1.6.12
+
+- Fix ignoring the apim.managedServiceAccount in deployment files
```
jgiovaresco, line 7:
```suggestion
- Fix ignoring the cockpit.managedServiceAccount in deployment files
```

jgiovaresco (commented):
You also need to update the artifacthub.io/changes in https://github.com/gravitee-io/helm-charts/pull/236/files#diff-f5a41f430a49651ab46e6bc027bc1e864d7af49fe3e9364b674b35b62476b74fR23

Also, I saw you removed the `serviceAccountName` from cockpit UI deployment. I don't know if it was intended. I'm really not comfortable with this service account thing... I don't when it should be defined (because maybe the cockpit generator also don't need it) 🤔

PR #238, opened 2022-04-11
chore: jetty config AM chart

remisultan:
**Issue**

https://github.com/gravitee-io/issues/issues/7479

**Description**

Every possible configurable jetty configuration in AM is now parameterized in the AM helm chart

**Additional context**

### DIFF
--- a/am/templates/api/api-configmap.yaml
+++ b/am/templates/api/api-configmap.yaml
@@ -12,9 +12,32 @@ metadata:
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
 data:
   gravitee.yml: |
-    {{- if .Values.api.ssl.enabled }}
     jetty:
-      secured: true
+      host: {{ .Values.api.http.host | default "0.0.0.0" }}
+      port: {{ .Values.api.http.port | default 8093 }}
+      idleTimeout: {{ .Values.api.http.idleTimeout | default 30000 }}
+      acceptors: {{ .Values.api.http.acceptors | default -1 }}
+      selectors: {{ .Values.api.http.selectors | default -1 }}
+      statistics: {{ .Values.api.http.statistics | default false }}
+      outputBufferSize: {{ .Values.api.http.outputBufferSize | default 32768 }}
+      requestHeaderSize: {{ .Values.api.http.requestHeaderSize | default 8192 }}
+      responseHeaderSize: {{ .Values.api.http.responseHeaderSize | default 8192 }}
+      {{- if .Values.api.http.pool }}
+      pool:
+        minThreads: {{ .Values.api.http.pool.minThreads | default 10 }}
+        maxThreads: {{ .Values.api.http.pool.maxThreads | default 200 }}
+        idleTimeout: {{ .Values.api.http.pool.idleTimeout  | default 60000 }}
+        queueSize: {{ .Values.api.http.pool.queueSize | default 6000 }}
+      {{ end }}
+      {{- if .Values.api.http.accesslog }}
+      accesslog:
+        enabled: {{ .Values.api.http.accesslog.enabled | default true }}
+        {{- if .Values.api.http.path }}
+        path: {{ .Values.api.http.accesslog.path }}
+        {{ end }}
+      {{ end }}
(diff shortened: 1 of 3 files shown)


am/values.yaml
```diff
@@ -236,17 +236,37 @@ api:
...
-  #    type: jks # Supports jks, pem, pkcs12
-  #    path: ${gravitee.home}/security/keystore.jks
-  #    password: secret
-    clientAuth: false
-  #  truststore:
-  #    type: jks # Supports jks, pem, pkcs12
-  #    path: ${gravitee.home}/security/truststore.jks
-  #    password: secret
+  jetty:
+    secured: true
```
jhaeyaert, line 240: Are you sure that the `secured` flag is `true` by default ?

am/templates/api/api-configmap.yaml
```diff
@@ -12,25 +12,50 @@ metadata:
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
 data:
   gravitee.yml: |
-    {{- if .Values.api.ssl.enabled }}
+    {{- if .Values.api.jetty }}
```
jhaeyaert, line 15: I'm wondering how this change will work on an existing installation ?
Does it mean that users using `api.ssl.enabled: true` must switch to `api.jetty.secured: true` ?

jhaeyaert (changes requested):

am/values.yaml
remisultan: my bad. I assumed it was per default 🙈

am/templates/api/api-configmap.yaml
remisultan: In this case yes.

tcompiegne: Agree with @jhaeyaert , we should keep consistency with all others products. 

I'm still not sure if we have to use : `Values.api.jetty.*` rather than `Values.api.http.*` 🤔 

@jhaeyaert what do you think ?

tcompiegne (changes requested):

am/templates/api/api-configmap.yaml
jhaeyaert: Yeah I think keeping `Values.api.http.*` is better as we do not introduce a thigh coupling with the underlying server we are using for our api (ie: jetty). Also, it avoid introducing breaking changes so...

remisultan: @jhaeyaert @tcompiegne fair enough

I'll put the values under `Values.api.http.*` and I'll keep using the `ssl.enabled` flag

am/templates/api/api-configmap.yaml
```diff
@@ -12,9 +12,33 @@ metadata:
...
+        queueSize: {{ .Values.api.http.pool.queueSize | default 6000 }}
+      {{ end }}
+      {{- if .Values.api.http.accesslog }}
+      accesslog:
+        enabled: {{ .Values.api.http.accesslog.enabled | default true }}
+        {{- if .Values.api.http.path }}
+        path: {{ .Values.api.http.accesslog.path }}
+        {{ end }}
+      {{ end }}
+      secured: {{ .Values.api.http.secured | default false }}
```
tcompiegne, line 40: `secured` should be kept in the `{{- if .Values.api.ssl.enabled }}` section (with value true).

am/templates/api/api-configmap.yaml
```diff
@@ -12,9 +12,33 @@ metadata:
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
 data:
   gravitee.yml: |
-    {{- if .Values.api.ssl.enabled }}
+    {{- if .Values.api.http }}
     jetty:
```
tcompiegne, line 16: `jetty` section should always be available, because if a customer uses the `Values.api.ssl.enabled` before, it will not work anymore

tcompiegne (changes requested):

am/templates/api/api-configmap.yaml
remisultan: FP

jhaeyaert (approved):

PR #242, opened 2022-04-14
feat(ae): Add support for Technical API

brasseld:


ae/values.yaml
```diff
@@ -186,4 +186,38 @@ engine:
...
+  services:
+    core:
+      http:
+        enabled: true
+        port: 18072
+        host: localhost
+        authentication:
+          type: basic
+          password: adminadmin
+        secured: false
```
kamiiiel, line 200: Is this used anywhere?

ae/values.yaml
```diff
@@ -186,4 +186,38 @@ engine:
...
+        ssl:
+          keystore:
+            type: "PKCS12"
+            path: "/p12/keystore"
```
kamiiiel, line 204: Is this used anywhere?

kamiiiel, line 204: Is this used anywhere?

ae/templates/engine-deployment.yaml
```diff
@@ -102,6 +102,10 @@ spec:
           ports:
             - name: {{ .Values.engine.service.internalPortName }}
               containerPort: {{ .Values.engine.service.internalPort }}
+            {{- if .Values.engine.services.core.http.enabled }}
+            - name: {{ .Values.engine.name }}-techapi
+              containerPort: {{ .Values.engine.services.core.http.port | default "18072" }}
```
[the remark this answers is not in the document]
brasseld: Yes, was done for keeping coherency with other products

ae/values.yaml
brasseld: Copy and paste :) will remove it

ae/templates/engine-service.yaml
```diff
@@ -24,6 +24,12 @@ spec:
     - port: 5701
       name: {{ .Values.engine.name }}-hz
       protocol: TCP
+    {{- if .Values.engine.services.core.service.enabled }}
+    - port: {{ .Values.engine.services.core.service.externalPort }}
+      targetPort: {{ .Values.engine.services.core.http.port }}
+      protocol: TCP
+      name: {{ .Values.engine.name }}-techapi
+    {{- end }}
```
jhaeyaert, line 32: I don't get this part 🧐
What is does mean `.Values.engine.services.core.service.enabled` ? Which "service" is enabled here ? Should it be `.Values.engine.services.core.service.http.enabled` instead ?

Same for `.Values.engine.services.core.service.externalPort` ?

jhaeyaert: Ok it refers to the "core" service which corresponds to the technical api itself. The "http" is only a way to expose the technical api (it is the only supported for now). So makes sense to keep it like this and keep consistency with other products (apim, am). 👍

jhaeyaert (approved):

PR #245, opened 2022-04-19
fix: Reorder HPA resources to avoid outofsync state with ArgoCD

brasseld:


jgiovaresco (approved):
🤩  we also notice that issue with 🥒

PR #247, opened 2022-04-19
feat: Add support for PodDisruptionBudget

brasseld:
To have some context: https://innablr.com.au/blog/what-is-pod-disruption-budget-in-k8s-and-why-to-use-it/

### DIFF
--- a/cockpit/templates/api/api-pdb.yaml
+++ b/cockpit/templates/api/api-pdb.yaml
@@ -1,0 +1,25 @@
+{{- if and (.Values.api.enabled) (.Values.api.pdb.enabled) -}}
+apiVersion: {{ include "common.capabilities.policy.apiVersion" . }}
+kind: PodDisruptionBudget
+metadata:
+  name: {{ template "gravitee.api.fullname" . }}
+  labels:
+    app.kubernetes.io/name: {{ template "gravitee.name" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/version: {{ .Values.api.image.tag | default .Chart.AppVersion | quote }}
+    app.kubernetes.io/component: "{{ .Values.api.name }}"
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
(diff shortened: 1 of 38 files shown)


cockpit/templates/api/api-pdb.yaml
```diff
@@ -0,0 +1,25 @@
...
+    app.kubernetes.io/component: "{{ .Values.api.name }}"
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+spec:
+  {{- if .Values.api.pdb.maxUnavailable }}
+  maxUnavailable: {{ .Values.api.pdb.maxUnavailable }}
+  {{ end -}}
+  {{- if .Values.api.pdb.minAvailable }}
+  minAvailable: {{ .Values.api.pdb.minAvailable }}
+  {{ end -}}
```
jgiovaresco, line 19: Shouldn't we drive the content with the Values?
Other we would need to upgrade the chart to support more settings 🤔 

```
 {{ toYaml .Values.api.pdb | nindent 4 -}}
```

cockpit/values.yaml
```diff
@@ -193,6 +193,11 @@ api:
     periodSeconds: 30
     failureThreshold: 3
 
+  pdb:
+    enabled: false
+    minAvailable: ""
+    maxUnavailable: "50%"
```
jgiovaresco, line 199: Do we need to set values if we disable pdb by default?

cockpit/templates/api/api-pdb.yaml
brasseld: I would say why not, but PDB is very straightforward, not a lot you can do https://docs.openshift.com/container-platform/4.9/rest_api/policy_apis/poddisruptionbudget-policy-v1.html

cockpit/values.yaml
brasseld: Good question, I left it there for providing example...

jgiovaresco: We could maybe put a comment with a link to the doc 🤷

jgiovaresco (approved):

PR #250, opened 2022-04-20
[AM] - Allow to use a k8s secret as mongodb uri

famaridon:
**Issue**

https://github.com/gravitee-io/issues/issues/7540

**Description**

Allow to use a secret as mongodb uri

**Additional context**

### DIFF
--- a/am/README.adoc
+++ b/am/README.adoc
@@ -89,6 +89,14 @@ URI].
 |`+mongo.uri+` |Mongo URI |`+null+`
 |===
 
+You can externalize mogodb uri in a k8s secret
+[cols=",,",options="header",]
+|===
+|Parameter |Description |Default
+|`+mongo.uriSecret.name+` |k8s secrect containing the uri |`+null+`
+|`+mongo.uriSecret.key+` |k8s secrect key containing the uri |`+null+`
+|===
+
 If no `+mongo.uri+` is provided, you can provide a `+mongo.servers+` raw
 definition in combination with `+mongo.dbname+`, plus eventual
 authentication configuration:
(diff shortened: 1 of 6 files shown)


leleueri:
Hi @famaridon 

Thanks your for you proposal. At a first look it is good (I didn't test it yet). 
But I'm wondering if this change is necessary as since AM 3.15 we are able to read secret from kubernetes.

You simply have to define you applicaiton settings like this `kubernetes://<namespace>/secrets/<my-secret-name>/<my-secret-key>`. Also you have to create the ServiceAccount that allow AM to connect to the Kubernetes API (the helm chart should do it by default)

Here is an example for my mongodb uri:

```
mongo:
  uri: kubernetes://default/secrets/mongo/mongouri
```

 Can you confirm that this way to define secret is working on your environment?

famaridon:
Hello this is a nice feature and it works. Please document it in the helm chart :).

But (because their is always a 'but') to do this I have updated the helm chart from 1.0.29 to the latest 1.0.41. And now you create a ClusterRole to configure RBAC but the ClusterRole is currently not namespaced.
But currently I'm using namespaces to deploy multiple environments. And I'm getting error on the second deployement.

`Error: UPGRADE FAILED: rendered manifests contain a resource that already exists. Unable to continue with update: ClusterRole "moovapps-future-am-cluster-role" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal "demo": current value is "testing"`

famaridon:
I used am.nameOverride as a workaround. and all works fine.

famaridon:
All is ok with the `kubernetes://default/secrets/mongo/mongouri` syntax.
I think you can abort this pull request you solution works fine

Could you add some notes in k8s chart documentation about this syntax ?

leleueri:
> Could you add some notes in k8s chart documentation about this syntax ?

Yes I will do https://github.com/gravitee-io/issues/issues/7630

PR #253, opened 2022-05-05
feat: Add support for managed ServiceAccounts name provided by user

kamiiiel:
**Issue**

https://github.com/gravitee-io/issues/issues/XXXXX

**Description**

A small description of what you did in that PR.

**Additional context**

### DIFF
--- a/ae/CHANGELOG.md
+++ b/ae/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 This file documents all notable changes to [Gravitee.io Alert Engine](https://github.com/gravitee-io/helm-charts/tree/master/ae) Helm Chart. The release numbering uses [semantic versioning](http://semver.org).
 
+### 1.1.35
+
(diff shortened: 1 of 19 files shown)


jgiovaresco (approved):

leleueri (changes requested):
Can you please add Unittest for AM and provide documentation about different options ?

brasseld (approved):

kamiiiel:
> Can you please add Unittest for AM and provide documentation about different options ?

I added the test for AM as well and regarding the documentation, maybe it is better to put this in a page or something, this is a settings similar for all products. What do you think @brasseld ?

brasseld:
> > Can you please add Unittest for AM and provide documentation about different options ?
> 
> I added the test for AM as well and regarding the documentation, maybe it is better to put this in a page or something, this is a settings similar for all products. What do you think @brasseld ?

Good point @kamiiiel I'm just wondering what is the best place for this.

PR #255, opened 2022-05-09
APIM: fix for #7140 - introduce ingressClassName parameter

HerrmannHinz:
**Issue**

https://github.com/gravitee-io/issues/issues/7140

**Description**

introduce `ingressClassName` as optional parameter/value for APIM 3.x helm chart 

**Additional context**

see related issue for detailed context

### DIFF
--- a/apim/3.x/Chart.yaml
+++ b/apim/3.x/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 name: apim3
 # When the version is modified, make sure the artifacthub.io/changes list is updated
 # Also update CHANGELOG.md
-version: 3.1.43
+version: 3.1.44
 appVersion: 3.15.8
 description: Official Gravitee.io Helm chart for API Management 3.x
 home: https://gravitee.io
--- a/apim/3.x/Chart.yaml
+++ b/apim/3.x/Chart.yaml
@@ -20,4 +20,4 @@ annotations:
   # List of changes for the release in artifacthub.io
   # https://artifacthub.io/packages/helm/graviteeio/apim3?modal=changelog
   artifacthub.io/changes: |
-    - Disable automatic download of Redis plugin
+    - Introduce `ingressClassName` as optional value/parameter
--- a/apim/3.x/values.yaml
+++ b/apim/3.x/values.yaml
@@ -523,6 +523,7 @@ api:
             password: adminadmin
         ingress:
           enabled: false
+          #ingressClassName: nginx
 #          path: /management/_(.*)
 #          hosts:
 #            - apim.example.com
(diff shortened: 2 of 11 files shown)


HerrmannHinz:
@brasseld @ytvnr  could u plz have a look? or assign ping the relevant persons to have a look? thanks in ahead :D

apim/3.x/values.yaml
```diff
@@ -585,6 +586,7 @@ api:
         kubernetes.io/ingress.class: nginx
         nginx.ingress.kubernetes.io/configuration-snippet: "etag on;\nproxy_pass_header ETag;\nproxy_pass_header if-match;\n"
         # kubernetes.io/tls-acme: "true"
+      #ingressClassName: nginx
```
brasseld, line 589: As `nginx` is already the default value for the `kubernetes.io/ingress.class` annotation, I think it's probably better to uncomment this line (and the others). 

wdyt ?

brasseld (changes requested):
Can you please add some unit tests ?

apim/3.x/values.yaml
HerrmannHinz: yeah. lets comment the `kubernetes.io/ingress.class: nginx` annotation. 
== `ingressClassName: nginx` should be commented in instead?

HerrmannHinz:
> Can you please add some unit tests ?

will do. never wrote helm unit tests  but let me give it a try.

apim/3.x/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 name: apim3
 # When the version is modified, make sure the artifacthub.io/changes list is updated
 # Also update CHANGELOG.md
-version: 3.1.43
+version: 3.1.44
```
kamiiiel, line 5: line 23 also needs to be updated

apim/3.x/values.yaml
kamiiiel: As it is already mentioned in the ticket, you can NOT have both the annotation and ingressClassName together but from what I see you just added the ingressClassName and without doing something about the "kubernetes.io/ingress.class: nginx" annotation or maybe I missed something?

apim/3.x/Chart.yaml
HerrmannHinz: done

apim/3.x/values.yaml
germainlefebvre4: IngressClass is THE way to define Ingress objects to the Ingress Controller on currently supported kubernetes versions.

Indeed annotations for ingress class has to be commented to allow old k8s versions to be supported and to be compliant with the end of deprecation (=removal) for ingress class annotation in 1.24+

No need to make a switch depending on the version because it would confuse parameters settings.

brasseld:
Closed by https://github.com/gravitee-io/helm-charts/pull/267

PR #261, opened 2022-05-20
feat(apim): add gravitee-license secret to apim's chart

mariano-fernandez:
**Issue**

https://github.com/gravitee-io/gravitee-cockpit/issues/2476

**Description**

To be able to fully automate apim-ee provisioning, the secret containing the license needs to be also created with the official chart

### DIFF
--- a/apim/3.x/CHANGELOG.md
+++ b/apim/3.x/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 This file documents all notable changes to [Gravitee.io API Management 3.x](https://github.com/gravitee-io/helm-charts/tree/master/apim/3.x) Helm Chart. The release numbering uses [semantic versioning](http://semver.org).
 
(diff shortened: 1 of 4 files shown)


brasseld (commented):
I'm not a big fan of that sort of "enterprise" change.
Also, starting from next APIM / AM version, it will be possible to provide base64 license from environment variables.
So I think we can forget this PR for the time being.

mariano-fernandez:
> I'm not a big fan of that sort of "enterprise" change. Also, starting from next APIM / AM version, it will be possible to provide base64 license from environment variables. So I think we can forget this PR for the time being.

@brasseld to be able to fully automate provisioning with helm, we thought that this could be the cleaner solution. Maybe we can have a quick call to discuss how to move forward without handling "enterprise" specific requirements ?

PR #265, opened 2022-06-10
feat(console): Add support for APIM console Pendo analytics with helm charts

ThibaudAV:
**Issue**
https://github.com/gravitee-io/issues/issues/7781

**Description**

Add support for APIM console Pendo analytics with helm charts

**Additional context**

### DIFF
--- a/apim/3.x/templates/api/api-configmap.yaml
+++ b/apim/3.x/templates/api/api-configmap.yaml
@@ -88,6 +88,10 @@ data:
         url: "https://{{index .Values.api.ingress.management.hosts 0 }}{{ .Values.api.ingress.management.path }}"
       ui:
         url: "https://{{index .Values.ui.ingress.hosts 0 }}{{ regexFind "/[a-zA-Z]+" .Values.ui.ingress.path }}"
+      {{- if (((.Values.ui).analytics).pendo).enabled }}
+      analytics:
+        {{ toYaml .Values.ui.analytics | indent 8 | trim -}}
+      {{- end }}
     services:
       core:
         http:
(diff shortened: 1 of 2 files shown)


apim/3.x/templates/api/api-configmap.yaml
```diff
@@ -88,6 +88,12 @@ data:
         url: "https://{{index .Values.api.ingress.management.hosts 0 }}{{ .Values.api.ingress.management.path }}"
       ui:
         url: "https://{{index .Values.ui.ingress.hosts 0 }}{{ regexFind "/[a-zA-Z]+" .Values.ui.ingress.path }}"
+      {{- if (((.Values.ui).analytics).pendo).enabled }}
+      analytics:
+        pendo:
+          enabled: {{ .Values.ui.analytics.pendo.enabled }}
+          apiKey: {{ .Values.ui.analytics.pendo.apiKey }}
```
Okhelifi, line 95: Do you think it's better to replace the content in analytics by something like this ?

```suggestion
 {{ toYaml .Values.ui.analytics | indent 8 }}
```

ThibaudAV: I don't really have an opinion, it makes the possible configurations less explicit but more flexible. 
As you wish ? ;)

Okhelifi (approved):

apim/3.x/templates/api/api-configmap.yaml
Okhelifi: As there is no real rule and not some much lines we can let it like it is. :)

PR #269, opened 2022-06-17
feat: store configuration with sensitive data in secrets

sebastien-helbert:
**Issue**

https://github.com/gravitee-io/issues/issues/7845

**Description**

Gravitee helm charts must store some configuration files in "Secrets" instead of "ConfigMap" to keep sensitive data safe.


gaetanmaisse:
@sebastien-helbert sorry for the late feedback on that... First, thank you for your contribution! 🎉 

After an internal discussion, this looks to be a good improvement however we won’t be able to merge it right now as it will be a major change for our users. On our side, we are working on refactoring the `gravitee.yml` handling, the idea is to offer multiple solutions without breaking the actual process. So I think we can close this PR for now and maybe recheck it when the changes we’ve in progress are merged.

PR #275, opened 2022-06-28
feat: add networking.k8s.io/v1 ingress apiVersion to apim 1.x

jbouchery:
**Issue**

https://github.com/gravitee-io/issues/issues/7925

**Description**

Include `networking.k8s.io/v1` ingress apiVersion in Gravitee APIM 1.X helm chart

### DIFF
--- a/apim/1.x/templates/_helpers.tpl
+++ b/apim/1.x/templates/_helpers.tpl
@@ -58,9 +58,11 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 Return the apiVersion of ingress.
 */}}
 {{- define "ingress.apiVersion" -}}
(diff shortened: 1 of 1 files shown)


brasseld:
Hello @jbouchery 

We are no more supporting apim 1.x for a long time now, and so the same rule is applied to the APIM 1.x Helm Charts.
I'm wondering if you can rely on Chart 3.x and update the tags for deploying APIM 1.x.
I'll be honest with you, that's not something I tried before telling you...

Regards,

jbouchery:
Hello @brasseld 

It was a feature request to quick fix an ingress for a stuck GKE upgrade. The upgrade to APIM 3.x wan't planed so soon but it seems to be the safest action to to :)

Regards

PR #281, opened 2022-08-03
feat(am): add template to enable SSl connections with Backends

leleueri:
related-to gravitee-io/issues#5856

### DIFF
--- a/am/CHANGELOG.md
+++ b/am/CHANGELOG.md
@@ -5,6 +5,7 @@ This file documents all notable changes to [Gravitee.io Access Management 3.x](h
 ### 1.0.46
 
 - [X] Upgrade Mongodb and Elasticsearch dependencies
+- [X] Add settings for Backend SSL connection
 
 ### 1.0.45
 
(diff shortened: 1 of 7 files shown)


ashraf706 (approved):

famaridon:
Is there documentation on how to use it?

leleueri:
After looking at the documentation, I amended the documentation for RDMS backends but not for Mongo :( 
I will update the doc and will detail the usage for HelmChart.

Basically, you have to enable the ssl link in the value.yaml: 
https://github.com/gravitee-io/helm-charts/blob/master/am/values.yaml#L48

Then you have to configure the keystore/truststore section https://github.com/gravitee-io/helm-charts/blob/master/am/values.yaml#L61 using extra volumes mounted from secrets. Here after an extra volume example for [the EE license key](https://docs.gravitee.io/am/current/am_installguide_kubernetes_ee.html)

```
  extraVolumeMounts: |
    - name: am-license
      mountPath: /opt/graviteeio-am-management-api/license
      readOnly: true
  extraVolumes: |
    - name: am-license
      secret:
        secretName: am-license
```
I have reopened https://github.com/gravitee-io/issues/issues/7729

PR #282, opened 2022-08-10
Authorize some special characters when fetching UI ingress path to compute management.url setting

phiz71:
**Issue**

https://github.com/gravitee-io/issues/issues/8303

**Description**

authorize some special characters when fetching UI ingress path to compute management.url setting


gaetanmaisse (approved):

gaetanmaisse:
Need to update apim/3.x/Chart.yaml too

PR #289, opened 2022-08-22
Added attributes in values.yaml and modified flow in templates to eas…

olaf-meyer:
…e rollout of gravitee in an OpenShift cluster

**Issue**

No issue

**Description**

During the deployment of Gravitee on OpenShift for IGS, we encountered two major problems:
1. OpenShift requires to start the pods with a "unique" user id, for this we needed to explicit overwrite the securityContext attribute for each deployment in the customes yaml file for APIM, AM and AE.
2. The ingress traffic in OpenShift is based on routes that are usually created from Ingress resources. This is not working at the moment because Ingress resource define a nginx ingress class that doesn't exist on most of the OpenShift cluster (most customer use the default ingress)

With this pull request I address these problems. Could you please verify if the changes are okay, then I'm going to modify the AM and AE as well.

### DIFF
--- a/apim/3.x/templates/api/api-deployment.yaml
+++ b/apim/3.x/templates/api/api-deployment.yaml
@@ -90,7 +90,7 @@ spec:
       initContainers:
         {{- if and .Values.jdbc.driver (eq .Values.management.type "jdbc") }}
         - name: get-repository-jdbc-ext
-          {{- toYaml .Values.initContainers | nindent 10 }}
+          {{- include "common.initcontainer.securitycontext" (dict "initContainers" .Values.initContainers "openshift" .Values.openshift) | indent 8}}
           command: ['sh', '-c', "mkdir -p /tmp/plugins-ext && cd /tmp/plugins-ext && wget  {{ .Values.jdbc.driver }}"]
           volumeMounts:
             - name: graviteeio-apim-repository-jdbc-ext
@@ -105,7 +105,7 @@ spec:
         - name: {{ template "gravitee.api.fullname" . }}
           image: "{{ .Values.api.image.repository }}:{{ default .Chart.AppVersion .Values.api.image.tag }}"
           imagePullPolicy: {{ .Values.api.image.pullPolicy }}
-          securityContext: {{ toYaml ( .Values.api.securityContext | default .Values.api.deployment.securityContext ) | nindent 12 }}
+          securityContext: {{include "common.container.securitycontext" (dict "currentSecurityContext" .Values.api.securityContext "defaultSecurityContext" .Values.api.deployment.securityContext "openshift" .Values.openshift)| nindent 12}}
           ports:
             - name: {{ .Values.api.service.internalPortName }}
               containerPort: {{ .Values.api.service.internalPort }}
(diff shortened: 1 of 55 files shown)


apim/3.x/templates/api/api-deployment.yaml
```diff
@@ -90,7 +90,7 @@ spec:
       initContainers:
         {{- if and .Values.jdbc.driver (eq .Values.management.type "jdbc") }}
         - name: get-repository-jdbc-ext
-          {{- toYaml .Values.initContainers | nindent 10 }}
+          {{- include "common.initcontainer.securitycontext" (dict "initContainers" .Values.initContainers "openshift" .Values.openshift) | indent 8}}
```
brasseld, line 93: I'm wondering why this section is needed as you can already override the securityContext section from the values.yaml.
See https://github.com/gravitee-io/helm-charts/blob/master/apim/3.x/values.yaml#L1276

I'm not sure we really want a different behavior because of OpenShift here.

Do I miss something?

brasseld (changes requested):
Hello @olaf-meyer 

Sounds generally a great contribution, thanks for this!
I just have some concern regarding the securityContext for initContainers as I wonder why we can't just rely on the existing values.yaml and why we do want  a different behavior for OpenShift.

apim/3.x/templates/api/api-deployment.yaml
olaf-meyer: Hello @brasseld,

let me first describe the problem how I have encountered it:
OpenShift want to start each container  with a "random" userid and groupid. This will happen if the container is not started with an explicit userid and/or groupid. This applies init-container and "normal" container in a pod. As an user you can define the userid and groupid. In most environment a platform admin needs to allow this.

In the current version of the yaml file, there a couple of default values for userId and groupId (for each deployment and for the init-container) that I needed to overwrite in order to run it on OpenShift. . My changes would only affect the current default setting, if the attribute openshift.enabled is true. This sets default values for all places where a security context defined a userid and/or groupid. If I leave out the changes for the initContainer, then in order to run gravitee on OpenShift, I need to set the attribute openshift.enabled and overwrite the securityContext for the initContainer.

I hope this clarifies my intentions.

brasseld: Probably a misunderstanding on my side, but at the end, why this:

```yaml
initContainers:
  image: alpine:latest
  imagePullPolicy: Always
  securityContext:
    runAsUser: 1001
    runAsNonRoot: true
  env: []

openshift:
  enabled: false
  securityContext:
    runAsUser: null
    runAsGroup: null
    runAsNonRoot: true
  ingress:
    generateRoute: true
    annotations:
      route.openshift.io/termination: "edge"
```
Could not be just something like this (for init containers)

```yaml
initContainers:
  image: alpine:latest
  imagePullPolicy: Always
  securityContext:
    runAsUser: null
    runAsGroup: null
    runAsNonRoot: true
  env: []

openshift:
  enabled: false
  ingress:
    generateRoute: true
    annotations:
      route.openshift.io/termination: "edge"
```

I don't really understand why we need a dedicated openshift securityContext section, and why we are not reusing the securityContext for initContainers or for the different pods we are deploying.

At the end, I sense that only the ingress should be managed in an other way for OpenShift, and the securityContext can be entirely managed with the existing values.yaml

PR #291, opened 2022-08-22
Bugs/portnametolong

bmixte:
**Issue**

to be created ?
https://github.com/gravitee-io/issues/issues/

**Description**

Following the following error during a test deployment.
helm-controller  reconciliation failed: Helm upgrade failed: failed to create resource: Deployment.apps "abc-d-01--abcd-efg" is invalid: spec.template.spec.containers[0].ports[1].name: Invalid value: "abcd-efg-techapi": must be no more than 15 characters
To solve it, I just truncate port name for bridge and technical port on api and gateway deployment templates.

**Additional context**

We are trying to deploy 2 control plane on the same cluster and must rename the release, deployments and pods.


bmixte:
@brasseld you are right.
I will fix this.

PR #292, opened 2022-08-22
fix: error port name too long

bmixte:
**Issue**

[#8345](https://github.com/gravitee-io/issues/issues/8345)

**Description**

Following the following error during a test deployment.  
helm-controller reconciliation failed: Helm upgrade failed: failed to create resource: Deployment.apps "abc-d-01--abcd-efg" is invalid: spec.template.spec.containers[0].ports[1].name: Invalid value: "abcd-efg-techapi": must be no more than 15 characters  
To solve it, I just truncate port name for bridge and technical port on api and gateway deployment templates.

**Additional context**

We are trying to deploy 2 control plane on the same cluster and must rename the release, deployments and pods.

### DIFF
--- a/apim/3.x/templates/api/api-deployment.yaml
+++ b/apim/3.x/templates/api/api-deployment.yaml
@@ -110,11 +110,11 @@ spec:
             - name: {{ .Values.api.service.internalPortName }}
               containerPort: {{ .Values.api.service.internalPort }}
             {{- if .Values.api.http.services.core.http.enabled }}
-            - name: {{- printf "%s-%s" .Values.api.name "techapi" | trunc 15 }}
+            - name: {{ printf "%s-%s" (.Values.api.name | trunc 8 | trimSuffix "-") "techapi" }}
               containerPort: {{ .Values.api.http.services.core.http.port | default "18083" }}
             {{- end }}
             {{- if .Values.api.services.bridge.enabled }}
-            - name: {{- printf "%s-%s" .Values.api.name "bridge" | trunc 15 }}
+            - name: {{ printf "%s-%s" (.Values.api.name | trunc 8 | trimSuffix "-") "bridge" }}
               containerPort: {{ .Values.api.services.bridge.service.internalPort }}
             {{- end }}
           env:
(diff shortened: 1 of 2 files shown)


apim/3.x/templates/api/api-deployment.yaml
```diff
@@ -110,11 +110,11 @@ spec:
             - name: {{ .Values.api.service.internalPortName }}
               containerPort: {{ .Values.api.service.internalPort }}
             {{- if .Values.api.http.services.core.http.enabled }}
-            - name: {{ .Values.api.name }}-techapi
+            - name: {{- printf "%s-%s" .Values.api.name "techapi" | trunc 15 }}
```
ThibaudAV, line 113:
```suggestion
            - name: {{ printf "%s-%s" .Values.api.name "bridge" | trunc 15 | trimSuffix "-" }}
```
seems to work better otherwise the test not work

📝 We have some note [here](https://github.com/gravitee-io/helm-charts/blob/master/apim/3.x/TESTS.md) to run test

ThibaudAV: if we add a trim here like this. it is possible that the ports have the same name if the api name is long. As this will remove the suffix `techapi`

WDYT about something like this :
`{{ printf "%s-%s" (.Values.api.name | trunc 8 | trimSuffix "-") "techapi" }}`

ThibaudAV:
Hello, and thank you for your contribution 🙂

apim/3.x/templates/api/api-deployment.yaml
bmixte: Sounds good to me.
I will update the code and push branch.

ThibaudAV (approved):
LGTM. 
But i wait an @gravitee-io/archi review. 

FI, I would do the merge in an intermediate branch and then in master to validate the CI.
I would add in addition some unit tests 


<details>
  <summary>Like this in `deployment_test.yaml`</summary>

  ```yml
  ...
  - it: Deploy with long api name
    template: api/api-deployment.yaml
    set:
      api:
        name: "I-am-a-very-long-name"
        services:
          bridge:
            enabled: true
            service:
              externalPort: 92
              internalPort: 18092
    asserts:
      - hasDocuments:
          count: 1
      - isKind:
          of: Deployment
      - equal:
          path: spec.template.spec.containers[0].ports
          value: 
            - containerPort: 8083
              name: http
            - containerPort: 18083
              name: I-am-a-techapi
            - containerPort: 18092
              name: I-am-a-v-bridge
  ```
</details>

brasseld:
LGTM but looking at the unit tests now.

PR #294, opened 2022-08-24
Add a startup probe on m-api

NicolasGeraud:
**Description**

Depending on the time to execute java upgraders during a version upgrade, we encountered issues when k8s restarted the pod even if the upgrade was not finished.
The "workaround" is to add a startup probe to accept a long initial startup.


brasseld (changes requested):
Can this be replicated to all charts, and add missing uni tests ?

NicolasGeraud:
unit tests added.
The startup latency is due to java upgrader so I don't think it makes sense to add a startup probe to other components.

brasseld (approved):

gaetanmaisse (approved):

gaetanmaisse:
⚠️ We need to update the version and changelog before releasing this

gaetanmaisse:
We will work on https://github.com/gravitee-io/helm-charts/pull/296 today and merge then release them both

PR #296, opened 2022-08-25
fix(apim): error port name too long

ThibaudAV:
**Issue**

https://github.com/gravitee-io/issues/issues/8345

Followup of https://github.com/gravitee-io/helm-charts/pull/292

**Description**

Crop all ports names to ensure they have a correct size

**Additional context**

### DIFF
--- a/am/templates/api/api-service.yaml
+++ b/am/templates/api/api-service.yaml
@@ -31,7 +31,7 @@ spec:
       {{ end }}
       {{ end }}
       {{- end }}
-      name: {{ .Values.api.name }}
+      name: {{ .Values.api.name | trunc 63 | trimSuffix "-" }}
     {{- if .Values.api.http.services.core.service.enabled }}
     - port: {{ .Values.api.http.services.core.service.externalPort }}
       targetPort: {{ .Values.api.http.services.core.http.port }}
@@ -43,7 +43,7 @@ spec:
       appProtocol: http
       {{ end }}
       {{- end }}
-      name: {{ .Values.api.name }}-technical
+      name: {{ printf "%s-%s" (.Values.api.name | trunc 53 | trimSuffix "-") "technical" }}
     {{- end }}
   selector:
     app.kubernetes.io/name: {{ template "gravitee.name" . }}
--- a/am/templates/gateway/gateway-service.yaml
+++ b/am/templates/gateway/gateway-service.yaml
@@ -31,7 +31,7 @@ spec:
       {{ end }}
       {{ end }}
       {{- end }}
-      name: {{ .Values.gateway.name }}
+      name: {{ .Values.gateway.name | trunc 63 | trimSuffix "-" }}
     {{- if .Values.gateway.services.core.service.enabled }}
     - port: {{ .Values.gateway.services.core.service.externalPort }}
       targetPort: {{ .Values.gateway.services.core.http.port }}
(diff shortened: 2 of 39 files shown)


mariano-fernandez (approved):

leleueri (commented):
Can you add unit tests on AM please and add an entry into the Changelogs + Chart.yml to inform about the change ?
In addition, name limitation look to be restricted to the POD spec, not in the sevice spec, but for consistency against products, I'm ok to apply it on AM too.

mariano-fernandez:
> Can you add unit tests on AM please and add an entry into the Changelogs + Chart.yml to inform about the change ?

There is already a unit-test for cockpit, but can we get the `changelog + chart.hml` part ? 😬

ThibaudAV:
@leleueri I didn't really find any doc on this subject.
But if it's only for pods then I leave as before on services ? or limit to 63
And add limit only when needed

am/templates/gateway/gateway-service.yaml
```diff
@@ -31,7 +31,7 @@ spec:
       {{ end }}
       {{ end }}
       {{- end }}
-      name: {{ .Values.gateway.name }}
+      name: {{ .Values.gateway.name | trunc 15 | trimSuffix "-" }}
```
leleueri, line 34: On Service resource, names are limited to 63 chars

remisultan:
ditto all of the above for AE with `Unit tests + Charts.yml`  change

am/templates/gateway/gateway-service.yaml
```diff
@@ -43,7 +43,7 @@ spec:
       appProtocol: http
       {{ end }}
       {{- end }}
-      name: {{ .Values.gateway.name }}-technical
+      name: {{ printf "%s-%s" (.Values.gateway.name | trunc 5 | trimSuffix "-") "technical" }}
```
leleueri, line 46: On Service resource, names are limited to 63 chars so here the trunc should be done at 53 chars in order to be able to append "-technical"

am/templates/api/api-service.yaml
```diff
@@ -31,7 +31,7 @@ spec:
       {{ end }}
       {{ end }}
       {{- end }}
-      name: {{ .Values.api.name }}
+      name: {{ .Values.api.name | trunc 15 | trimSuffix "-" }}
```
leleueri, line 34: On Service resource, names are limited to 63 chars

am/templates/api/api-service.yaml
```diff
@@ -43,7 +43,7 @@ spec:
       appProtocol: http
       {{ end }}
       {{- end }}
-      name: {{ .Values.api.name }}-technical
+      name: {{ printf "%s-%s" (.Values.api.name | trunc 5 | trimSuffix "-") "technical" }}
```
leleueri, line 46: On Service resource, names are limited to 63 chars so here the trunc should be done at 53 chars in order to be able to append "-technical"

am/templates/ui/ui-service.yaml
```diff
@@ -27,7 +27,7 @@ spec:
       appProtocol: http
       {{ end }}
       {{- end }}
-      name: {{ .Values.ui.name }}
+      name: {{ .Values.ui.name | trunc 15 | trimSuffix "-" }}
```
leleueri, line 30: On Service resource, names are limited to 63 chars

leleueri (changes requested):

ThibaudAV:
Following discussion with @leleueri . The limit is kept at 63 for services and 15 deployment. as the one imposed by k8s

leleueri (approved):

gaetanmaisse (approved):

PR #298, opened 2022-09-13
rollingUpdate maxUnavailable set to 25% to avoid stop when only 1 pod…

famaridon:
**Description**

Set default rolling update max unavailable to 25% (the default value) instead of 1 to avoid stop and restart if only 1 pod is available.

It's not a bug but it's a trap :).


leleueri (approved):

leleueri:
Thanks you for your contribution :+1:

PR #299, opened 2022-09-13
[APIM] Fix missing config to configure email into gravitee.yml

ThibaudAV:
**Issue**

https://graviteesource.zendesk.com/agent/tickets/4792 

**Description**

Fix missing config to configure email into gravitee.yml

**Additional context**


gaetanmaisse (approved):

gaetanmaisse:
Reverting as it leads to a weird config map if the old attribute is used:
image

PR #300, opened 2022-09-13
Revert "[APIM] Fix missing config to configure email into gravitee.yml"

gaetanmaisse:
Reverts gravitee-io/helm-charts#299


gaetanmaisse:
Self-merging to restore a correct behavior, see https://github.com/gravitee-io/helm-charts/pull/299#issuecomment-1245646591

PR #301, opened 2022-09-13
APIM: Use `smtp.properties` directly instead of copying properties one by one

gaetanmaisse:
**Description**

Reapply https://github.com/gravitee-io/helm-charts/pull/299 with a breaking change to ensure the generated config-map is ok.

BREAKING CHANGE: Use `smtp.properties.starttls.enable` instead of `smtp.properties.starttlsEnable`


gaetanmaisse:
Need to double-check with @ThibaudAV

PR #304, opened 2022-09-20
feat(ae): add optional license value to enable usage of alert engine

passionne:
**Issue**

N/A

**Description**

Add an optional value `engine.secrets.licensekey` in alert engine `values.yml` that create a kubernetes secret used by AE deployment.

**Additional context**

N/A


ae/templates/engine-secrets.yaml
```diff
@@ -0,0 +1,11 @@
+{{- with .Values.engine.secrets }}
```
brasseld, line 1: Can you please follow the same "convention" as the one we had for the cockpit certificates ?

https://github.com/gravitee-io/helm-charts/blob/master/apim/3.x/templates/cockpit/cockpit-certificates.yaml#L1

brasseld: So basically, having the license not at the engine level, but at the root level.
Also, this is interesting for AE, but that would be awesome to replicate this for APIM / AM

brasseld (changes requested):

ae/templates/engine-secrets.yaml
AleksandarGjorgievski: Can we also take into consideration that we can use the same secret for AE, AM and APIM as long as they are in the same namespace?
So we could check if the secret already exists before we create it.

passionne:
> Can you confirm why you want a TechOps review on this change? I've asked around and it seems that's not something we usually do. Did you mean DevOps (@sebdevaux) perhaps?
> 
> FWIW the change looks reasonable to me, but I'm not an expert on our helm charts or helm in general.

Sorry, It's my first PR on this repo and I wanted to be sure that everything is ok. No mandatory review for Techops here 😊.

ae/templates/engine-secrets.yaml
passionne: Yes good remarks, and thanks for your feedback, I am working on it.

passionne: ✔ I let you keep a fresh eyes on it.
Thank you !

brasseld (approved):

PR #311, opened 2022-11-02
fix: Replace ClusterRole with Role

kamiiiel:
This PR is to solve this issues https://community.gravitee.io/t/unable-to-install-am-helm-chart-in-more-than-one-namespace-in-the-same-cluster/1248/1

### DIFF
--- a/am/CHANGELOG.md
+++ b/am/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 This file documents all notable changes to [Gravitee.io Access Management 3.x](https://github.com/gravitee-io/helm-charts/tree/master/am/) Helm Chart. The release numbering uses [semantic versioning](http://semver.org).
 
+### 1.0.51
+
+- [X] Replace ClusterRole with Role
+
 ### 1.0.50
 
 - [X] Manage mfa_rate limit on gateway
--- a/am/Chart.yaml
+++ b/am/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 name: am
 # When the version is modified, make sure the artifacthub.io/changes list is updated
 # Also update CHANGELOG.md
-version: 1.0.50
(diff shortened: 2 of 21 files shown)


jgiovaresco (approved):

brasseld:
@leleueri Can you please check that one ?

am/templates/common/clusterrole.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.am.managedServiceAccount }}
+{{- if and .Values.am.managedServiceAccount .Values.am.clusterRoleRules }}
```
leleueri, line 1: Do we still need clusterRole manifest ?

am/templates/common/clusterrolebinding.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.am.managedServiceAccount }}
+{{- if and .Values.am.managedServiceAccount .Values.am.clusterRoleRules }}
```
leleueri, line 1: Do we still need clusterRoleBinding manifest ?

leleueri (changes requested):

am/templates/common/clusterrole.yaml
kamiiiel: We don't need it if the user wants to put everything in the same namespace but it is better to keep the cluster role option for those users who want to:
1) Have access to different resources in other namespaces
2) Resolve property files inside gravitee.yaml using KubernetesPropertyResolver from all namespaces

kamiiiel:
It can be `managedServiceAccount: false` by deafult

am/templates/common/clusterrole.yaml
leleueri: in that case ok.

PR #318, opened 2022-12-14
Adding note about roles and namespaces

AndrewMcFarlandCampbellGravitee:
**Issue**

https://graviteedevops.atlassian.net/browse/DOC-293

**Description**

I added a note about using secrets and namespaces to the APIM and AM `README.*` files. The `adocc` files are `include`d in pages on the docs site. The note was added to the `md` files as well to keep them in sync.

### DIFF
--- a/am/README.adoc
+++ b/am/README.adoc
@@ -111,7 +111,7 @@ mongo:
 ====
 If you need to access a secret, you have to create a role within your namespace.
 
-If you need to access a secret in another namespace, you have to create a role in that namespace. The two roles can have the same name, but they are completely seperate objects.
+If you are deploying in another namespace and you need to access a secret there, you have to create a separate role in that namespace. The two roles can have the same name, but they are completely separate objects - each role only gives access to the namespace it is created in.
 ====
 
 ==== Shared configuration
--- a/am/README.adoc
+++ b/am/README.adoc
@@ -107,6 +107,13 @@ mongo:
   uri: kubernetes://default/secrets/mongo/mongouri
 ----
 
+[TIP]
+====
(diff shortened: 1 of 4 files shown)


am/README.adoc
```diff
@@ -107,6 +107,13 @@ mongo:
   uri: kubernetes://default/secrets/mongo/mongouri
 ----
 
+[TIP]
+====
+If you need to access a secret, you have to create a role within your namespace.
+
+If you need to access a secret in another namespace, you have to create a role in that namespace. The two roles can have the same name, but they are completely seperate objects.
```
ivan-gravitee, line 114:
```suggestion
If you are deploying in another namespace and you need to access a secret there, you have to create a separate role in that namespace. The two roles can have the same name, but they are completely separate objects - each role only gives access to the namespace it is created in.
```

sebdevaux: The suggested change are fine for me. Do we have to give the needed parameter to create this role or this have to go in an other part of the documentation? 
```
apim:
  managedServiceAccount: true
  roleRules:
    - apiGroups:
        - ""
      resources:
        - configmaps
        - secrets
      verbs:
        - get
        - list
        - watch
```

ivan-gravitee: In my view if this is a helm/k8s thing we should just reference the proprietary documentation, e.g. https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-example

WDYT?

ivan-gravitee:
Thanks @AndrewMcFarlandCampbellGravitee, this looks good. 
@sebdevaux is it worth mentioning that we don't support `ClusterRole`?

ivan-gravitee (approved):

PR #319, opened 2022-12-15
chore(apim): complete alert engine configuration

remisultan:
**Issue**

https://github.com/gravitee-io/issues/issues/8733

**Description**

This PR brings additional configuration from the alert engine connector

### DIFF
--- a/apim/3.x/templates/api/api-configmap.yaml
+++ b/apim/3.x/templates/api/api-configmap.yaml
@@ -457,6 +457,7 @@ data:
         enabled: true
         ws:
           discovery: true
+          {{ toYaml .Values.alerts.options | nindent 10 | trim -}}
           {{- with .Values.alerts.endpoints }}
           endpoints:
             {{ toYaml . | nindent 12 | trim -}}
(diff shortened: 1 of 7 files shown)


gaetanmaisse:
Looks good! Would like to have helm master's opinion on this also, 👋🏻 @sebdevaux

apim/3.x/templates/api/api-configmap.yaml
```diff
@@ -457,6 +457,16 @@ data:
         enabled: true
         ws:
           discovery: true
+          sendEventsOnHttp: {{ .Values.alerts.sendEventsOnHttp | default true }}
```
sebdevaux, line 460: I would prefer a block of code to handle this part. 
It's more flexible.
WDYT?

apim/3.x/templates/gateway/gateway-configmap.yaml
```diff
@@ -360,6 +360,16 @@ data:
         enabled: true
         ws:
           discovery: true
+          sendEventsOnHttp: {{ .Values.alerts.sendEventsOnHttp | default true }}
```
sebdevaux, line 363: I would prefer a block of code to handle this part. 
It's more flexible.
WDYT?

sebdevaux (approved):

PR #338, opened 2023-02-16
Add `labels` in the metrics config of the Management API and Gateway

gaetanmaisse:
**Issue**

https://github.com/gravitee-io/issues/issues/7927

**Description**

Add `labels` in the metrics config of the Management API and Gateway

### DIFF
--- a/apim/3.x/tests/api/configmap_metrics_test.yaml
+++ b/apim/3.x/tests/api/configmap_metrics_test.yaml
@@ -1,0 +1,48 @@
+suite: Test Management API default configmap - Metrics
+templates:
+  - "api/api-configmap.yaml"
+tests:
+  - it: Disable metrics service by default
+    template: api/api-configmap.yaml
+    asserts:
+      - notMatchRegex:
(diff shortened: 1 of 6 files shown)


apim/3.x/tests/api/configmap_metrics_test.yaml
```diff
@@ -0,0 +1,42 @@
+suite: Test Management API default configmap - Metrics
+templates:
+  - "api/api-configmap.yaml"
+tests:
+  - it: Enable metrics service and set `labels`
```
jgiovaresco, line 5: Maybe we could add a test case for the default config (without setting anything just checking the values.yaml)

jgiovaresco (approved):

apim/3.x/tests/api/configmap_metrics_test.yaml
gaetanmaisse: Good catch, just added!

jgiovaresco (approved):

phiz71 (approved):

PR #346, opened 2023-02-23
feat: add fullNameOverride to helpers

oleksiyp:
**Description**

Add `fullNameOverride` option to _helpers.tpl so that more flexibility is possible in naming such things `Service`


brasseld:
Hi @oleksiyp 

Thanks for the PR. Can you please add unit tests so we can check stuff properly ?

oleksiyp:
Ok

PR #348, opened 2023-02-27
feat(templates):  Add `externalTrafficPolicy` in service configuration

sebdevaux:
Add `externalTrafficPolicy` in service configuration

### DIFF
--- a/apim/3.x/templates/api/api-configmap.yaml
+++ b/apim/3.x/templates/api/api-configmap.yaml
@@ -345,9 +345,9 @@ data:
           {{- end }}
         {{- end }}
     # SMTP configuration used to send mails
+    {{- if .Values.smtp }}
     email:
       enabled: {{ .Values.smtp.enabled }}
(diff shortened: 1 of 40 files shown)


apim/3.x/templates/api/api-configmap.yaml
```diff
@@ -345,9 +345,9 @@ data:
           {{- end }}
         {{- end }}
     # SMTP configuration used to send mails
+    {{- if .Values.smtp.enabled }}
```
gaetanmaisse, line 348: Wondering it we should not rely on `.Values.smtp` only, in case the APIM admin wants to disable the smtp and blocks its activation in the UI (would be `email.enabled: false`)

gaetanmaisse: Also, can you add the Jira + GH Issues links in the commit message 😇

gaetanmaisse (approved):

PR #352, opened 2023-03-14
logback.xml

sebdevaux:
**Issue**

Unable to use the value parameter `logging.debug: true` when using an external gravitee.yml

**Description**

This pr allow user to set `logging.debug: true` when using an external gravitee.yml:
- a configMag is generated with logback.xml
- The provided gravitee.yml is used by the application


gaetanmaisse (commented):
Would like a 2nd reviewer or a live pair review with you @sebdevaux  because I'm pretty light on the subject 😁

phiz71 (approved):

PR #370, opened 2023-04-18
Improve gateway probes

sebdevaux:
BREAKING CHANGE

Refactor probes configuration:

- Move probe configuration under `deployment:``
- Remove `apisync` parameter on gateway readinessProbe 
- Improve Gateway probes by using exposed probes (/_node/health?probes=http-server /_node/health?probes=api-sync)

At startup we are now waiting for sync. 
Related documentation:
https://gravitee.slab.com/posts/apim-gateway-probes-good-practices-jzb9yghl

@brasseld FYI


phiz71 (commented):
LGTM. I have 2 questions though:
 - should we define default probes pour `api` also?
 - maybe I confuse the concepts but I would have applied `api-sync` on the readiness probe and `http-server` on the startup. Because the gateway can be started but not ready yet to accept calls because all APIs have not been synchronized.

phiz71 (approved):

PR #372, opened 2023-04-18
feat(cockpit): notification definition added

ashraf706:


sebdevaux (changes requested):
Upgrade the Chart version and update the changelog please

sebdevaux (approved):

PR #403, opened 2023-06-12
feat(cockpit): add onboarding incomplete configuration in cockpit

aurelien-pacaud:
**Issue**

https://gravitee.atlassian.net/browse/COC-231

**Description**

Add a new property to set time before notifying when user incomplete his onboarding

### DIFF
--- a/cockpit/tests/api/api-configmap_gravitee_yaml_test.yaml
+++ b/cockpit/tests/api/api-configmap_gravitee_yaml_test.yaml
@@ -101,7 +101,8 @@ tests:
                 starttls.enable : "false"
             
             onboarding:
-              notify-uncompleted-cron: 0 */30 * * * *
+              incomplete:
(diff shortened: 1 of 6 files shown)


cockpit/tests/api/api-configmap_gravitee_yaml_test.yaml
```diff
@@ -1462,12 +1471,13 @@ tests:
                 responseType: code
                 scopes: openid,profile,email
 
-  - it: should configure onboarding to 5 mins
+  - it: should configure to notifying incomplete onboarding 30 mins later
```
passionne, line 1474: As 30 mins is the default value, we may use another value to check the difference.
WDYT ?

aurelien-pacaud: Good point ! Change to 10 mins

PR #472, opened 2024-10-25
feat: spec-gen helm

remisultan:
This PR introduces the necessary values for spec-gen helm charts https://gravitee.atlassian.net/browse/SAT-133

### DIFF
--- a/cockpit/Chart.yaml
+++ b/cockpit/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 name: cockpit
 # When the version is modified, make sure the artifacthub.io/changes list is updated
 # Also update CHANGELOG.md
-version: 2.0.1
+version: 2.1.0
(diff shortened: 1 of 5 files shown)


cockpit/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 name: cockpit
 # When the version is modified, make sure the artifacthub.io/changes list is updated
 # Also update CHANGELOG.md
-version: 2.0.1
+version: 2.1.0
 appVersion: 4.46.0
```
aurelien-pacaud, line 6: The version should be updated too?

remisultan: ?
```suggestion
appVersion: 4.51.0
```

aurelien-pacaud (approved):

